PR #4, opened 2019-08-05
add update user endpoint

naytis:


backend/app/Http/Controllers/Api/Auth/AuthController.php
```diff
@@ -0,0 +1,38 @@
...
+use Illuminate\Http\JsonResponse;
+use App\Http\Controllers\Controller;
+
+final class AuthController extends Controller
+{
+    private $updateUserAction;
+
+    public function __construct(UpdateUserAction $updateUserAction)
+    {
+        $this->middleware('auth:jwt', ['except' => []]);
```
lenchvolodymyr: move middleware to route/api.php

lenchvolodymyr: probably better to group auth routes

lenchvolodymyr: https://laravel.com/docs/5.8/routing#route-groups

backend/app/Http/Controllers/Api/Auth/AuthController.php
```diff
@@ -0,0 +1,38 @@
...
+
+        $this->updateUserAction = $updateUserAction;
+    }
+
+    public function update(UpdateUserHttpRequest $request): JsonResponse
+    {
+        $response = $this->updateUserAction->execute(
+            new UpdateUserRequest(
```
lenchvolodymyr: Add static method public static function fromRequest(UpdateUserHttpRequest $request) { return new static(...); }

backend/app/Http/Requests/UpdateUserHttpRequest.php
```diff
@@ -0,0 +1,25 @@
...
+use Illuminate\Foundation\Http\FormRequest;
+
+final class UpdateUserHttpRequest extends FormRequest
+{
+    public function authorize(): bool
+    {
+        return true;
+    }
+
+    public function rules(): array
```
lenchvolodymyr: it is fine, but could you also add "get" methods with type hintings it will prevent some bugs in the future:

public function getName() { return $this->get('name'); }

backend/app/Repositories/UserRepository.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Repositories;
+
+use App\Entities\User;
+use App\Repositories\Contracts\UserRepository as IUserRepository;
+
+final class UserRepository implements IUserRepository
```
lenchvolodymyr: Name repository as EloquentUserRepository, such as we are going to have also ElasticSearchRepository. It is best practice not to call implementation different from interface

backend/routes/api.php
```diff
@@ -16,3 +16,9 @@
 Route::middleware('auth:api')->get('/user', function (Request $request) {
     return $request->user();
 });
+
+Route::prefix('v1')->group(function () {
+    Route::group(['prefix' => 'auth', 'namespace' => 'Api\\Auth'], function () {
+        Route::put('/update', 'AuthController@update');
```
lenchvolodymyr: it is incorrect name of route, REST doesn't have verbs in the resource name. For update users route should be

PUT /api/v1/users

backend/app/Actions/User/UpdateUserAction.php
```diff
@@ -0,0 +1,36 @@
...
+
+    public function execute(UpdateUserRequest $request): UpdateUserResponse
+    {
+        try {
+            $user = $this->userRepository->getById($request->id());
+        } catch (ModelNotFoundException $exception) {
+            throw new UserNotFoundException();
+        }
+
+        $user->email = $request->email() ?: $user->email;
```
lenchvolodymyr: add to your getters default value and pass there $user->email, it reads much better

public function getEmail(string $default = '') : string { return $this->email ?: $default; }

lenchvolodymyr: $request->getEmail($user->email)

backend/app/Actions/User/UpdateUserRequest.php
```diff
@@ -0,0 +1,57 @@
...
+            $request->password()
+        );
+    }
+
+    public function getId(): int
+    {
+        return $this->id;
+    }
+
+    public function getName(string $default): ?string
```
lenchvolodymyr: the only thing "default" shouldn't be required

PR #6, opened 2019-08-06
add-jwt-library

mes1901:
jwt library with route and endpoint


backend/app/Http/Controllers/Auth/RegisterController.php
```diff
@@ -4,69 +4,30 @@
...
-     * @return \Illuminate\Contracts\Validation\Validator
-     */
-    protected function validator(array $data)
+    public function create(Request $request)
     {
-        return Validator::make($data, [
-            'name' => ['required', 'string', 'max:255'],
-            'email' => ['required', 'string', 'email', 'max:255', 'unique:users'],
-            'password' => ['required', 'string', 'min:8', 'confirmed'],
+        $validator = Validator::make($request->all(), [
```
TatsuUkraine: create separate Http request object and define validation there

backend/app/Http/Controllers/Auth/RegisterController.php
```diff
@@ -4,69 +4,30 @@
...
-    protected function create(array $data)
-    {
-        return User::create([
-            'name' => $data['name'],
-            'email' => $data['email'],
-            'password' => Hash::make($data['password']),
+        if ($validator->fails()) {
+            return response()->json($validator->errors());
+        }
+        $user = User::create([
```
TatsuUkraine: create action service and move all business logic there, controllers should be slim

backend/app/Http/Controllers/Auth/RegisterController.php
```diff
@@ -1,72 +1,38 @@
...
-    | Register Controller
-    |--------------------------------------------------------------------------
-    |
-    | This controller handles the registration of new users as well as their
-    | validation and creation. By default this controller uses a trait to
-    | provide this functionality without requiring any additional code.
-    |
-    */
-
     use RegistersUsers;
```
lenchvolodymyr: Get rid of all default code and invoke your own action that should create a user, with your request and response.

backend/app/Http/Requests/RegisterHttpRequest.php
```diff
@@ -0,0 +1,24 @@
...
+
+declare(strict_types=1);
+
+namespace App\Http\Requests;
+
+use Illuminate\Foundation\Http\FormRequest;
+
+final class RegisterHttpRequest extends FormRequest
+{
+    public function authorize()
```
lenchvolodymyr: add to request getters. e.g. getName(): string, getEmail(): string

backend/app/Repositories/UserRepository.php
```diff
@@ -0,0 +1,16 @@
...
+declare(strict_types=1);
+
+namespace App\Repositories;
+
+use App\User;
+use App\Repositories\Contracts\UserRepository as IUserRepository;
+
+final class UserRepository implements IUserRepository
+{
+    public function create(array $fields): User
```
lenchvolodymyr: name method save, and pass there User entity, that will be persisted and returned, that is save(User $user) { $user->save(); return $user; }

backend/app/Actions/User/RegisterUserAction.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+declare(strict_types=1);
+
+namespace app\Actions\User;
+
+use App\Repositories\UserRepository;
```
lenchvolodymyr: you bonded interface on implementation, that means you can use here contract.  App\Repositories\Contracts\UserRepository

backend/app/Repositories/UserRepository.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Repositories;
+
+use App\User;
+use App\Repositories\Contracts\UserRepository as IUserRepository;
+
+final class UserRepository implements IUserRepository
```
lenchvolodymyr: name implementation according to your persistence usage, e.g. EloquentUserRepository

backend/app/Actions/User/RegisterUserAction.php
```diff
@@ -0,0 +1,30 @@
...
+    public function execute(RegisterRequest $request): AuthenticationResponse
+    {
+        $user = $this->userRepository->create([
+            'name' => $request->getName(),
+            'email' => $request->getEmail(),
+            'password' => $request->getPassword()
+        ]);
+        $token = JWTAuth::fromUser($user);
+
+        return new AuthenticationResponse($token);
```
lenchvolodymyr: If you pass RegisterRequest it is expected you get RegisterResponse

backend/app/Actions/User/RegisterUserAction.php
```diff
@@ -0,0 +1,30 @@
...
+    private $userRepository;
+
+    public function __construct(UserRepository $userRepository)
+    {
+        $this->userRepository = $userRepository;
+    }
+
+    public function execute(RegisterRequest $request): RegisterResponse
+    {
+        $user = $this->userRepository->create([
```
lenchvolodymyr: you forgot change usage of repository according to method save. Create here new User model and assign properties from request

backend/app/Http/Controllers/Auth/RegisterController.php
```diff
@@ -1,72 +1,35 @@
...
-     */
-    protected function create(array $data)
+    protected function create(RegisterHttpRequest $request): JsonResponse
     {
-        return User::create([
-            'name' => $data['name'],
-            'email' => $data['email'],
-            'password' => Hash::make($data['password']),
-        ]);
+        $request = new RegisterRequest(
```
lenchvolodymyr: add static method fromHttpRequest(RegisterHttpRequest $request): RegisterRequest { return new static(/* your stuff here*/); }

and then call it RegisterRequest::fromHttpRequest($request)

backend/app/Http/Controllers/Auth/RegisterController.php
```diff
@@ -1,72 +1,35 @@
...
+            $request->get('name'),
+            $request->get('email'),
+            $request->get('password')
+        );
+        $response = $this->registerUserAction->execute($request);
+
+        return response()->json($response->getToken());
```
lenchvolodymyr: return token in our general response format

[
    "data" => [ "token" => $token ],
   "meta" => []
]

backend/app/Repositories/UserRepository.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Repositories;
+
+use App\User;
+use App\Repositories\Contracts\UserRepository as EloquentUserRepository;
+
+final class UserRepository implements EloquentUserRepository
```
lenchvolodymyr: not exactly, your interface was called correctly, because interface is general and does not depend on implementation. So when you use UserRepository you are not worried about where the user has been stored. But it is important when you binding implementaion to interface, because some repository can store your data via Eloquent framework, or in the files, or in the memory.

So rename class to EloquentUserRepository and interface leave as it was

backend/app/Http/Requests/RegisterHttpRequest.php
```diff
@@ -0,0 +1,39 @@
...
+    {
+        return false;
```
lenchvolodymyr: I am not sure about false

backend/app/Actions/User/RegisterUserAction.php
```diff
@@ -0,0 +1,33 @@
...
+
+    public function __construct(UserRepository $userRepository)
+    {
+        $this->userRepository = $userRepository;
+    }
+
+    public function execute(RegisterRequest $request): RegisterResponse
+    {
+        $user = $this->userRepository->save(
+            User::create([
```
lenchvolodymyr: if you call "create" the user will be stored to DB instantly. So you should create a new model ($user = new User;), then assignt properties $user->name = $request->getName(); and so on, and finally call save method of repository $userRepository->save($user);. Such actions are needed because of all actions you do with persisting data should be inside repositories

backend/app/Repositories/EloquentUserRepository.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Repositories;
+
+use App\User;
+use App\Repositories\Contracts\UserRepository as IUserRepository;
```
lenchvolodymyr: Now you don't need alias with I prefix

lenchvolodymyr:
pay attention that User now has namespace Entities\User

PR #8, opened 2019-08-06
add request-service (post/get wrapper over axios)

KoR82:


frontend/src/services/requestService.js
```diff
@@ -0,0 +1,68 @@
...
+    axios.put(url, {
+    	data: data,
+        headers: {
+            ...headers,
+            ...authHeader
+        }
+    });
+};
+
+const delete = (url, params = {}) => {
```
lenchvolodymyr: how did you manage to name variable "delete"? it is the reserved word isn't it?

KoR82: oops. How may be better: destroy or remove?

lenchvolodymyr: variable you can name as you wish, I mean destroy or deleteRequest. And then name key in the service as "delete", something like this
const requestDelete = () => {/* ...do stuff */}
export default {
    delete: requestDelete
};

but pay attention, according to airbnb style guide keys with values should be put on the top of an object, so if you have an update method, it will look like

export default {
    delete: requestDelete
    update
}

lenchvolodymyr: destroy is also okay =)

frontend/src/services/requestService.js
```diff
@@ -0,0 +1,68 @@
+import axios from 'axios';
+
+axios.defaults.headers.common = {
```
TatsuUkraine: don't modify global object, it's better to create axios instance, and define default params there

frontend/src/services/requestService.js
```diff
@@ -0,0 +1,68 @@
...
+    'X-Requested-With': 'XMLHttpRequest',
+    'Content-type': 'application/json',
+    'Accept': 'application/json'
+}
+
+const csrfToken = document.head.querySelector('meta[name="csrf-token"]');
```
TatsuUkraine: not sure if it's right way to get anything for axios params, such types of services shouldn't know anytthing about the DOM. Since if this service will be used in SSR it will fail

frontend/src/services/requestService.js
```diff
@@ -0,0 +1,68 @@
...
+
+if (csrfToken) {
+    axios.defaults.headers.common['X-CSRF-TOKEN'] = csrfToken.content;
```
TatsuUkraine: see first comment

frontend/src/services/requestService.js
```diff
@@ -0,0 +1,62 @@
+import axiosService from 'axios';
+
+const axios = axiosService.create {
```
lenchvolodymyr: what is this construction?

frontend/src/services/requestService.js
```diff
@@ -0,0 +1,62 @@
...
+    headers: {
+        'X-Requested-With': 'XMLHttpRequest',
+        'Content-type': 'application/json',
+        'Accept': 'application/json'
+    }
+};
+
+const token = 'getTokenFromAuthService';
```
lenchvolodymyr: token should be used dynamically. Use interceptors

https://github.com/axios/axios#interceptors

PR #11, opened 2019-08-06
Add registration form for frontend

devhackerone:
Form works, I'll add logo soon, will need to play around with the layout in the container


TatsuUkraine:
@devhackerone SIgnUp still uses style like `v-card` instead of `VCard` (https://vuejs.org/v2/style-guide/#Component-name-casing-in-templates-strongly-recommended). Components, that doesn't have any slot passed, not self-closing (https://vuejs.org/v2/style-guide/#Self-closing-components-strongly-recommended). Also props definition order same as was before (https://vuejs.org/v2/style-guide/#Element-attribute-order-recommended), v-model better put before any other props (except build in, see link), event binding better put last

Also, please, resolve any merge conflicts

frontend/src/components/auth/SignUp.vue
```diff
@@ -0,0 +1,141 @@
+<template>
+    <VContainer
+            fluid
```
TatsuUkraine: it seems you have too much spaces here and in other lines with props definition

PR #14, opened 2019-08-06
add web browser storage service

TumkoVolodymyr:


frontend/src/services/Storage.js
```diff
@@ -0,0 +1,52 @@
+const tokenKeyName = 'auth.access_token';
+class Storage
+{
```
TatsuUkraine: please keep code style according to JS styleguide, where is `{` should be on same line with classe, var or function definition

frontend/src/services/storage.js
```diff
@@ -0,0 +1,38 @@
...
+
+    hasToken() {
+        return !!this.get(tokenKeyName);
+    }
+
+    removeToken() {
+        return this.removeItem(tokenKeyName);
+    }
+}
+export const sessionStorage = new Storage('sessionStorage');
```
lenchvolodymyr: add method that will check if localStorage is accessible and create storage with this "driver". I mean

isLocalStorageAccessible = () => try { window.localStorage.setItem('test-local-storage', 1); window.localStorage.removeItem('test-local-storage'); return true; } catch(e) { return false; }

lenchvolodymyr: and then check let storage = new Storage(); if (!isLocalStorageAccessible()) { storage = new Storage('session'); } export default storage;

frontend/src/services/storage.js
```diff
@@ -0,0 +1,49 @@
...
+export const localStorage = new Storage();
+
+isLocalStorageAccessible = () => try {
```
lenchvolodymyr: you weren't supposed to copy my code)) but anyway, please, add const to function and use curly braces when you define function in this case, because it won't work. Get rid of exports of sessionStorage and localStorage variables.  And test it, please

PR #16, opened 2019-08-07
Feature/add jwt auth frontend

aborenko1996:


frontend/src/App.vue
```diff
@@ -1,28 +1,30 @@
 <template>
-  <div id="app">
-    <img alt="Vue logo" src="./assets/logo.png">
-    <HelloWorld msg="Welcome to Your Vue.js App"/>
-  </div>
+    <v-app id="app">
```
TatsuUkraine: use CamelCase format for component usage inside templates

frontend/src/App.vue
```diff
@@ -1,28 +1,30 @@
...
+        <top-navbar></top-navbar>
```
TatsuUkraine: use CamelCase format for component usage inside templates

frontend/src/App.vue
```diff
@@ -1,28 +1,30 @@
...
+        <router-view></router-view>
+    </v-app>
 </template>
 
 <script>
-import HelloWorld from './components/HelloWorld.vue'
+    import HelloWorld from './components/HelloWorld.vue';
```
TatsuUkraine: remove this component import, since you not using it

frontend/src/App.vue
TatsuUkraine: it's probably better add LayoutPage with base scaffold there (including top navbar) and use use nested routing to register it

frontend/src/components/auth/Login.vue
```diff
@@ -0,0 +1,149 @@
+<template>
+    <v-content>
+                <v-flex
```
TatsuUkraine: ensure that you have 4 spaces in each nested row here and in rest of the file

frontend/src/components/auth/Login.vue
```diff
@@ -0,0 +1,149 @@
...
+                                <v-text-field
+                                        name="email"
+                                        class="login-form-input"
+                                        v-model="email"
+                                        solo
+                                        type="text"
+                                        :rules="emailRules"
+                                        required
+                                >
+                                </v-text-field>
```
TatsuUkraine: consider using self closing tags where it's possible (see Vue Styleguide recomendation)

frontend/src/components/auth/Login.vue
```diff
@@ -0,0 +1,149 @@
...
+                    v => /.+@.+\..+/.test(v) || 'E-mail must be valid',
+                ],
+                passwordRules: [
+                    v => !!v || 'Password is required',
+                    v => (v && v.length >= 6) || 'Password must be equal or more than 6 characters'
+                ]
+            }
+        },
+        methods: {
+            ...mapActions({
```
TatsuUkraine: better to use namespaced map definition
```
...mapActions("auth", {
    "login"
})
```

frontend/src/components/navbar/TopNavbar.vue
```diff
@@ -0,0 +1,22 @@
...
+    <p v-if="!isLoggedIn">Not logged</p>
+    <p v-else>{{user.name}} <v-btn @click="logout">Logout</v-btn></p>
+</template>
+
+<script>
+    import {mapActions, mapGetters} from "vuex";
+
+    export default {
+        computed: {
+            ...mapGetters({
```
TatsuUkraine: use namespaced map definition (see prev comment)

frontend/src/store/modules/auth/actions.js
```diff
@@ -0,0 +1,30 @@
+export default {
+    login: (context, user) => {
```
TatsuUkraine: add folder "types" and define action const definitions there
```
export const LOGIN = "LOGIN";
```
and use them inside action definition and in components

TatsuUkraine: use explicit param definition where it's possible like, for example, instead of `context` better to use `{commit}`, since it's only one param that is used from context

frontend/src/store/modules/auth/getters.js
```diff
@@ -0,0 +1,6 @@
+export default {
+    hasToken: () => () => !!state.token,
```
TatsuUkraine: see comment about CONST usage in actions

frontend/src/store/modules/auth/mutations.js
```diff
@@ -0,0 +1,19 @@
+export default {
+    USER_LOGIN: (state, response) => {
```
TatsuUkraine: see comment about CONST usage in actions

PR #18, opened 2019-08-07
Feature/add endpoint for adding website

StasKozar:


backend/app/Http/Requests/AddWebsiteHttpRequest.php
```diff
@@ -0,0 +1,38 @@
...
+    {
+        return $this->get('name');
+    }
+
+    public function domain(): string
+    {
+        return $this->get('domain');
+    }
+
+    public function singlePage(): ?bool
```
lenchvolodymyr: it cannot be null as you validate it by boolean

StasKozar: Fixed.

backend/database/migrations/2019_08_07_114836_tracking_info_table.php
```diff
@@ -0,0 +1,21 @@
...
+
+use Illuminate\Support\Facades\Schema;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Database\Migrations\Migration;
+
+class TrackingInfoTable extends Migration
+{
+    public function up()
+    {
+        Schema::create('tracking_info', function (Blueprint $table) {
```
lenchvolodymyr: I cannot understand why did you create table only with one id? This table is useless because one web site has one tracking number. And number should be saved in seprate field. Add to the website field "tracking_number" and fill it incrementally programmatically

StasKozar: Fix

backend/app/Repositories/WebsiteRepository.php
```diff
@@ -0,0 +1,27 @@
...
+
+use App\Entities\Website;
+use App\Repositories\Contracts\EloquentWebsiteRepository;
+use Illuminate\Support\Facades\DB;
+
+final class WebsiteRepository implements EloquentWebsiteRepository
+{
+    public function save(Website $website): Website
+    {
+        $website->tracking_number = $this->getLastTrackingNumber() + 1;
```
lenchvolodymyr: This logic should be inside action

backend/app/Http/Controllers/Api/WebsiteController.php
```diff
@@ -0,0 +1,29 @@
...
+    {
+        $this->addWebsiteAction = $addWebsiteAction;
+    }
+
+    public function add(AddWebsiteHttpRequest $request): JsonResponse
+    {
+        $response = $this->addWebsiteAction->execute(
+            AddWebsiteRequest::fromRequest($request)
+        );
+
```
lenchvolodymyr: For http response use ApiResponse::success method

lenchvolodymyr: It meanse your response or resource should implement contract ApiResponse

backend/app/Providers/AppServiceProvider.php
```diff
@@ -17,6 +17,10 @@ public function register()
             \App\Repositories\Contracts\UserRepository::class,
             \App\Repositories\EloquentUserRepository::class
         );
+        $this->app->bind(
+            \App\Repositories\Contracts\EloquentWebsiteRepository::class,
```
lenchvolodymyr: Your interface is abstraction, your class should be detailed. You cannot abstract from eloquent because eloquent is particular implementation. So class should be eloquent, interface without eloquent

backend/app/Repositories/WebsiteRepository.php
StasKozar: Fix

backend/app/Http/Controllers/Api/WebsiteController.php
StasKozar: Fix

backend/app/Providers/AppServiceProvider.php
StasKozar: Fix

PR #36, opened 2019-08-08
Feature/dashboard

wyyi:


wyyi:
> Use camel case for component names in templates
it looks ugly as hell, but ok (☞ﾟヮﾟ)☞

TatsuUkraine:
It's not question of personal preference, it's question of best practices and Vue styleguide suggestions

TatsuUkraine:
@wyyi also remove node-sass from deps, and ensure that font definition is actually needed, if so it might be better move font definition to app or layout component, to ensure that it affects all app

frontend/src/plugins/router.js
```diff
@@ -11,34 +10,29 @@ Vue.use(Router);
 export default new Router({
     mode: 'history',
     base: '/',
-    routes: [
+    routes: [{
+            path: '',
+            name: 'home',
+            redirect: '/visitors'
```
TatsuUkraine: Use route name for redirect

frontend/src/components/Sidepanel.vue
```diff
@@ -0,0 +1,101 @@
...
+</template>
+
+<script>
+export default {
+    data: () => ({
+        drawer: true,
+        links: [{
+                icon: '/assets/icons/home.svg',
+                text: 'Home',
+                route: '/login'
```
TatsuUkraine: Use named route definitions if it's possible

frontend/src/components/Sidepanel.vue
```diff
@@ -0,0 +1,101 @@
+<template>
+    <VNavigationDrawer v-model="drawer" app clipped>
+        <VList class="mt-12" flat>
+            <VListItem class="pl-8 my-3" v-for="link in links" :key="link.text" router :to="link.route">
```
TatsuUkraine: If you have more than 3 attr in tag - define all of them with new line

frontend/src/components/Header.vue
```diff
@@ -0,0 +1,99 @@
+<template>
+    <VAppBar class="elevation-0 " clipped-left app>
+        <VLayout flex align-center class="pl-4 pr-10">
+            <VIcon class="mr-2 hidden-lg-and-up" @click="drawer = !drawer">menu</VIcon>
```
TatsuUkraine: Create method in component to toggle drawer prop

frontend/src/components/Header.vue
```diff
@@ -0,0 +1,99 @@
...
+                    </VListItem>
+                </VList>
+            </VMenu>
+        </VLayout>
+    </VAppBar>
+</template>
+
+<script>
+export default {
+    drawer: true,
```
TatsuUkraine: Don't define properties in component directly like you did here

frontend/src/components/Sidepanel.vue
```diff
@@ -0,0 +1,101 @@
+<template>
+    <VNavigationDrawer v-model="drawer" app clipped>
```
TatsuUkraine: You probably no need define v-model here and drawer prop, since you made drawer always opened, probably it's enough to use just `permanent` flag here

frontend/src/components/Navbar.vue
```diff
@@ -0,0 +1,28 @@
...
+    export default {
+        components: { Header, Sidepanel },
+        data: () => ({
+            drawer: true
+        })
+    }
+</script>
+
+<style lang="scss">
+    .v-navigation-drawer__border {
```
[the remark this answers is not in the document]
wyyi: whenever i try scoping this rule or moving it to the corresponding component, it stops affecting the target element, do you have any idea what could be the cause?

frontend/src/components/Sidepanel.vue
wyyi: should i consider the possibility of it being viewed on a mobile device and the drawer being closed by default?

frontend/src/components/Navbar.vue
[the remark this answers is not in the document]
TatsuUkraine: probably because you are trying to reach class name, that defined in child components https://vue-loader.vuejs.org/guide/scoped-css.html#child-component-root-elements you need to use `deep` or `::v-deep` selector in such cases

frontend/src/components/Sidepanel.vue
TatsuUkraine: you can, but it will require different approach how to handle open/close action for it, since open action (button) could be placed in another component, like header, so if you want to do this, you need to emit event from Header component, whenever menu button clicked (it can use custom v-model to inform it's parent), and this component will pass drawer prop to v-navigation component, also it should work in reverse direction

so if you want to do so - you can try, but it probably not required for this ticket

TatsuUkraine:
@wyyi  please resolve merge conflicts

TatsuUkraine:
@wyyi  please fix eslint issues

TatsuUkraine:
@wyyi please, fix eslint notes, PR won't be reviewed or merged before eslint errors and warnings will be fixed

PR #37, opened 2019-08-08
link login frontend with backend

naytis:


frontend/src/store/modules/auth/actions.js
```diff
@@ -1,9 +1,24 @@
...
-    [LOGIN]: (context, user) => {
-        return new Promise((resolve, reject) => {
+    [LOGIN]: async (context, user) => {
+        const url = process.env.VUE_APP_API_URL + '/auth/login';
+
+        try {
+            const response = await requestService.create(url, user);
+
+            context.commit(USER_LOGIN, response);
+            context.commit(SET_AUTHENTICATED_USER, user);
```
TatsuUkraine: User should be fetched from `me` endpoint, no need to save password value to state, it's unsecured

frontend/src/store/modules/auth/actions.js
```diff
@@ -1,9 +1,24 @@
...
 import {SET_AUTHENTICATED_USER, USER_LOGIN, USER_LOGOUT} from "./types/mutations";
+import requestService from "../../../services/requestService";
 
 export default {
-    [LOGIN]: (context, user) => {
-        return new Promise((resolve, reject) => {
+    [LOGIN]: async (context, user) => {
+        const url = process.env.VUE_APP_API_URL + '/auth/login';
+
+        try {
```
TatsuUkraine: It's better to use promise chain instead of try catch

frontend/src/store/modules/auth/actions.js
```diff
@@ -1,9 +1,24 @@
 import {LOGIN, LOGOUT} from './types/actions';
 import {SET_AUTHENTICATED_USER, USER_LOGIN, USER_LOGOUT} from "./types/mutations";
+import requestService from "../../../services/requestService";
 
 export default {
-    [LOGIN]: (context, user) => {
-        return new Promise((resolve, reject) => {
+    [LOGIN]: async (context, user) => {
+        const url = process.env.VUE_APP_API_URL + '/auth/login';
```
TatsuUkraine: Create config class with methods like getApiUrl, that will receive in constructor env params

Or define config file that exports functions like getApiUrl

In that way if necessary such file can be mocked for tests

frontend/src/store/modules/auth/actions.js
```diff
@@ -1,9 +1,24 @@
...
+
+        try {
+            const response = await requestService.create(url, user);
```
TatsuUkraine: It's probably better create folder api, create there file auth, that will expose functions like - authorize, logout etc, and use it to make API calls to backend

frontend/src/services/requestService.js
```diff
@@ -9,18 +10,14 @@ axiosInstance.interceptors.request.use((config) => {
...
+        config.headers.Authorization = 'Bearer ' + storage.getToken();
     }
 
     return config;
 });
 
-const getToken = () => {
-    return 'Bearer ' + 'tokenFromSessionStore';
-};
+axiosInstance.interceptors.response.use(response => response.data.data);
```
TatsuUkraine: Better remove this, or change just to `response.data` because otherwise no one will be able to get meta info from response

frontend/src/api/auth.js
```diff
@@ -0,0 +1,12 @@
+import requestService from "../services/requestService";
+import config from "../config";
+
+const resourceUrl = config.getApiUrl() + '/auth';
+
+const authorize = params => requestService.create(resourceUrl + '/login', params);
+const me = () => requestService.get(resourceUrl + '/me');
```
TatsuUkraine: Please rename this function to `getAuthUser`

frontend/src/store/modules/auth/actions.js
```diff
@@ -1,27 +1,25 @@
 import {LOGIN, LOGOUT} from './types/actions';
 import {SET_AUTHENTICATED_USER, USER_LOGIN, USER_LOGOUT} from "./types/mutations";
+import { authorize, me } from '../../../api/auth';
```
TatsuUkraine: Use alias @ instead of ../../..

frontend/src/store/modules/auth/actions.js
```diff
@@ -1,27 +1,25 @@
...
-                resolve(fakeUser);
-            }
+        return authorize(user)
+          .then(response => {
+            me()
+              .then(response => {
+                const user = response[0];
+                context.commit(SET_AUTHENTICATED_USER, user);
+              })
+              .catch(error => {
```
TatsuUkraine: Not sure that you need to catch and throw same error, this catch is redundant, same for other catch

frontend/src/store/modules/auth/actions.js
```diff
@@ -1,27 +1,25 @@
...
-            };
-
-            if(fakeUser.email === user.email && fakeUser.password === user.password) {
-                context.commit(USER_LOGIN, fakeUser);
-                context.commit(SET_AUTHENTICATED_USER, fakeUser);
-                resolve(fakeUser);
-            }
+        return authorize(user)
+          .then(response => {
+            me()
```
TatsuUkraine: To ensure that action will be finished when both requests finished, `then` should return result from me

frontend/src/store/modules/auth/mutations.js
```diff
@@ -1,8 +1,10 @@
 import {SET_AUTHENTICATED_USER, USER_LOGIN, USER_LOGOUT} from "./types/mutations";
+import storage from "../../../services/storage";
 
 export default {
     [USER_LOGIN]: (state, response) => {
-        state.token = response.access_token;
+        storage.setToken(response.token);
```
TatsuUkraine: Remove this row

lenchvolodymyr:
@naytis fix conflicts

frontend/.env.example
```diff
@@ -1 +1,2 @@
-VUE_APP_URL=localhost:8443
\ No newline at end of file
+VUE_APP_API_URL=http://<host>:<port>/api/v1
```
lenchvolodymyr: leave only VUE_APP_API_URL and get rid of VUE_APP_URL from request service

PR #41, opened 2019-08-08
Feature/add-reset-password-frontend

SerhiiPlotnikov:


lenchvolodymyr:
you added a few commits but looks like you added a whole application. Did you do hard reset?

SerhiiPlotnikov:
> you added a few commits but looks like you added a whole application. Did you do hard reset?

No, I have been writing my component, and then i decided to make pull request for a new state of develop branch. After that i made push request to my branch and got so much changed files

TatsuUkraine:
@SerhiiPlotnikov please fix merge conflicts

frontend/src/components/auth/ResetPasswordForm.vue
```diff
@@ -0,0 +1,167 @@
...
+                                required
+                        />
+                    </VForm>
+                </VCardText>
+                <VCardActions>
+                    <VBtn @click="onResetPassword" class="reset-password-form-button mt-3" color="#3C57DE">
+                        Send Password Reset Link
+                    </VBtn>
+                </VCardActions>
+                <v-alert class="status-success" v-if="status==='success'" type="success">
```
TatsuUkraine: use PascalCase style for component naming, to keep app style consistent

frontend/src/components/auth/ResetPasswordForm.vue
```diff
@@ -0,0 +1,167 @@
+<template>
+    <VContent>
+        <VFlex
+                lg6
```
TatsuUkraine: it seems you have unnecessary spaces here and in file in general

frontend/src/components/auth/ResetPasswordForm.vue
```diff
@@ -0,0 +1,167 @@
...
+                                v-model="email"
+                                solo
+                                type="text"
+                                :rules="emailRules"
+                                required
+                        />
+                    </VForm>
+                </VCardText>
+                <VCardActions>
+                    <VBtn @click="onResetPassword" class="reset-password-form-button mt-3" color="#3C57DE">
```
TatsuUkraine: ensure that your code uses styleguide suggestions, where even binding should places last one

frontend/src/components/auth/ResetPasswordForm.vue
TatsuUkraine: probably better avoid using v-if-else here. You can define computed prop like `showAlert` that will be true, when status not empty. Class value also can be move to computed prop and define it's value based on status, same for type

frontend/src/components/auth/ResetPasswordForm.vue
```diff
@@ -0,0 +1,167 @@
...
+                resetPassword: RESET_PASSWORD
+            }),
+            onResetPassword() {
+                if (this.$refs.form.validate()) {
+                    this.resetPassword({
+                        email: this.email,
+                    }).then(response => {
+                        this.status = 'success';
+                        this.message = response;
+                    }, err => {
```
TatsuUkraine: better to use `catch` chain method

frontend/src/pages/ResetPassword.vue
```diff
@@ -0,0 +1,18 @@
+<template>
+    <ResetPasswordForm></ResetPasswordForm>
```
TatsuUkraine: use self closing tags when it's possible

PR #43, opened 2019-08-09
Add master for adding websites

LymarMaryna:


frontend/src/components/website/adding_master/StepsProgressBar.vue
```diff
@@ -0,0 +1,80 @@
...
+        </VLayout>
+    </VContainer>
+</template>
+
+<script>
+    export default {
+        name: 'StepsProgressBar',
+        computed: {
+            stepNumber () {
+                return this.$parent.currentStepNumber;
```
TatsuUkraine: avoid using parent reference, component shouldn't know anything about it's parent, since it breaks single direction flow

frontend/src/components/website/AddWebsiteMaster.vue
```diff
@@ -0,0 +1,140 @@
...
+                </VLayout>
+            </VFlex>
+            <VFlex
+                lg6
+                md6
+                hidden-sm-and-down
+            >
+                <VImg
+                    :src="require('@/assets/running_man.jpg')"
+                >
```
TatsuUkraine: use self closing tags where it's possible

frontend/src/components/website/adding_master/StepAddDomain.vue
```diff
@@ -0,0 +1,98 @@
...
+        },
+        computed: {
+            errorText() {
+                return this.$parent.currentError;
+            }
+        },
+        methods: {
+            onGoToNextStep () {
+                if (this.$refs.form.validate()) {
+                    this.$emit('go-to-next-step', {
```
TatsuUkraine: it's probably better to use Vuex module to collect data between steps

frontend/src/components/website/adding_master/StepAddName.vue
```diff
@@ -0,0 +1,91 @@
...
+        },
+        computed: {
+            errorText() {
+                return this.$parent.currentError;
+            }
+        },
+        methods: {
+            onGoToNextStep () {
+                if (this.$refs.form.validate()) {
+                    this.$emit('go-to-next-step', {
```
TatsuUkraine: it's better to use Vuex module to collect data between steps

frontend/src/components/website/adding_master/StepTrackingInfo.vue
```diff
@@ -0,0 +1,77 @@
...
+                    </VCardText>
+                    <VCardText class="mb-2 pa-0">
+                        <div class="subtitle-2 mb-2 font-weight-bold">Website tracking</div>
+                        <div class="caption mb-2">This is Global Site Tag tracking code for your website.</div>
+                        <VTextarea
+                            solo
+                            auto-grow
+                            readonly
+                            rows="6"
+                            value="<!--Global site tag - Analitycs-->
```
TatsuUkraine: move this to computed prop, Vue template should contains as less logic as possible

frontend/src/pages/AddWebsitePage.vue
```diff
@@ -0,0 +1,14 @@
+<template>
+    <AddWebsiteMaster />
```
TatsuUkraine: it's probably doesn't make sense to create such page, since `AddWebsiteMaster` can be used as page (I mean move all from AddWebsiteMaster here and remove this `AddWebsitePage.vue`)

frontend/src/components/website/adding_master/StepAddDomain.vue
```diff
@@ -59,7 +59,11 @@
     export default {
         name: 'StepAddDomain',
         props: {
-            errorMessage: String,
+            errorMessage: {
+                type: String,
+                required: false,
```
TatsuUkraine: not a blocker, but required not required when you have default key)) (not a blocker, just something to remember for future)

frontend/src/components/website/adding_master/StepAddName.vue
```diff
@@ -47,7 +51,11 @@
     export default {
         name: 'StepAddName',
         props: {
-            errorMessage: String,
+            errorMessage: {
+                type: String,
+                required: false,
```
TatsuUkraine: see prev comment

frontend/src/components/website/adding_master/StepAddDomain.vue
```diff
@@ -0,0 +1,127 @@
...
+                    v => !!v || 'Website domain is required',
+                    v => domainRegex.test(v) || 'Domain must be correct'
+                ],
+            }
+        },
+        computed: {
+            ...mapGetters('website', {
+                newWebsite: GET_NEW_WEBSITE
+            }),
+            errorText() {
```
TatsuUkraine: no need to define computed prop for such cases, you can use prop directly in template

frontend/src/components/website/adding_master/StepAddDomain.vue
```diff
@@ -0,0 +1,127 @@
...
+        },
+        methods: {
+            ...mapActions('website', {
+                setDomain: SET_DOMAIN,
+                setSinglePage: SET_SPA,
+                saveNewSite: SAVE_NEW_WEBSITE
+            }),
+            onGoToNextStep () {
+                if (this.$refs.form.validate()) {
+                    this.setDomain(this.domain).then(() => {
```
TatsuUkraine: no need to define action per field, you can define 1 action and 1 mutation, like `SET_DOMAIN_INFO` to update multiple fields state, and SAVE_ACTION - for saving, and call them like
```
this.setDomainInfo(...).then(() => this.saveNewSite()).then(this.onSaved).catch(this.onError);
```
where `onSaved` - success behavior, and `onError` - error behavior

frontend/src/components/website/adding_master/StepsProgressBar.vue
```diff
@@ -0,0 +1,96 @@
...
+</template>
+
+<script>
+    export default {
+        name: 'StepsProgressBar',
+        props: {
+            stepNumber: {
+                type: Number,
+                required: true,
+                default: 1
```
TatsuUkraine: if you defined required: true, default is redundant

frontend/src/components/website/adding_master/StepsProgressBar.vue
```diff
@@ -0,0 +1,96 @@
+<template>
+    <VContainer>
+        <VLayout class="progress-panel">
+            <VFlex
+                ma-1
+                xs4
+                sm4
+                md4
+                class="progress-step"
+                :class="{'active': currentStep === 1,'filled': currentStep >= 1}"
```
TatsuUkraine: probably better define computed prop like "firstStepClass" or something like, that will return object based on currentStep

same for other step indicators (secondStepClass etc.)

frontend/src/components/website/adding_master/StepAddDomain.vue
```diff
@@ -0,0 +1,127 @@
...
+                        })
+                    });
+                }
+            },
+            onError (errors) {
+                if (errors.name) {
+
+                    this.$router.push({name: 'add_websites_step_1', params: {errorMessage: errors.name}});
+
+                } else if (errors.domain) {
```
TatsuUkraine: avoid using else if where it's possible, much cleaner implementation where `if` has empty return like
```
if (errors.name) {
....
    return;
}

if (errors.domain) {
.....
    return;
}

//etc.
```

PR #46, opened 2019-08-09
added endpoint for retrieving sessions

aborenko1996:


backend/app/Http/Controllers/Api/SessionController.php
```diff
@@ -0,0 +1,32 @@
...
+        $this->getAllSessionsAction = $getAllSessionsAction;
+    }
+
+    public function getAllSessions()
+    {
+        $response = $this->getAllSessionsAction->execute();
+
+        return ApiResponse::success(
+            new GetAllSessionsResponse([
+                'sessions' => SessionResource::collection($response->sessions())
```
TatsuUkraine: better to pass `$response` or `$response->sessions()` to constructor, and inside `toArray` format it like you need

backend/app/Repositories/Contracts/SessionRepository.php
```diff
@@ -0,0 +1,13 @@
...
+declare(strict_types=1);
+
+
+namespace App\Repositories\Contracts;
+
+use Illuminate\Support\Collection;
+
+interface SessionRepository
+{
+    public function all(): Collection;
```
TatsuUkraine: better rename method to `getCollection()`

backend/app/Repositories/SessionRepository.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Repositories;
+
+use App\Repositories\Contracts\SessionRepository as SessionRepositoryInterface;
+use Illuminate\Support\Collection;
+
+class SessionRepository implements SessionRepositoryInterface
```
TatsuUkraine: repository implementation should describe implementation itself, in our case it should be `EloquentSessionRepository`

backend/app/Http/Controllers/Api/SessionController.php
```diff
@@ -0,0 +1,31 @@
...
+        $this->getAllSessionsAction = $getAllSessionsAction;
+    }
+
+    public function getAllSessions()
+    {
+        $response = $this->getAllSessionsAction->execute();
+
+        return ApiResponse::success(
+            new GetAllSessionsResponse([
+                'sessions' => $response->sessions()
```
TatsuUkraine: you still passing array instead of object, pass `response` object or session collection, and format it inside `toArray`

TatsuUkraine: make it just `new GetAllSessionsResponse($response->sessions())` or `new GetAllSessionsResponse($response)` and convert to needed format in `toArray`, try to avoid using arrays for passing anything in constructors, that going to be used inside of it, since its responsibility of `GetAllSessionsResponse` to format response, and not controllers

backend/app/Http/Response/Sessions/GetAllSessionsResponse.php
```diff
@@ -0,0 +1,26 @@
...
+
+    public function __construct(Collection $sessionCollection)
+    {
+        $this->sessionCollection = $sessionCollection;
+    }
+
+    public function toArray(): array
+    {
+        return [
+            'sessions' => SessionResource::collection($this->sessionCollection)->jsonSerialize()
```
TatsuUkraine: `jsonSerialize` this method call probably is redundant here

PR #52, opened 2019-08-09
add new visitors endpoint

mes1901:


backend/app/Http/Controllers/Api/VisitorController.php
```diff
@@ -29,4 +36,15 @@ public function getAllVisitors()
             ])
         );
     }
+
+    public function getNewVisitors()
+    {
+        $response = $this->getNewVisitorsAction->execute();
+
+        return ApiResponse::success(
+            new GetNewVisitorsResponse([
```
lenchvolodymyr: Did you implement this response? It should return empty response, but work anyway. Write a test on this endpoint

mes1901: Response was checked and work well. I will write a test later

PR #58, opened 2019-08-09
fix middleware assignment on routes

naytis:


backend/routes/api.php
```diff
@@ -21,12 +21,12 @@
         Route::post('/register', 'AuthController@register');
         Route::post('/login', 'AuthController@login');
         Route::post('/reset-password', 'ResetPasswordController@sendPasswordResetLink');
-        Route::get('/me', 'AuthController@getCurrentUser')->middleware('auth:jwt');
+        Route::get('/me', 'AuthController@getCurrentUser')->middleware('auth:api');
```
lenchvolodymyr: Are you sure? Can you make a test? You can look at thread php, there is some abstraction for jwt that you can reuse

PR #65, opened 2019-08-10
restrict site content visibility to authenticated users

wyyi:


frontend/src/pages/Layout.vue
```diff
@@ -1,16 +1,29 @@
 <template>
     <VApp>
-        <Navbar />
-        <RouterView />
+        <Header :is-logged="isLogged" />
```
TatsuUkraine: with this approach you breaking single responsibility. Layout should be responsible just for layout, but not for auth guard

frontend/src/components/auth/LoginGuard.vue
```diff
@@ -1,16 +1,23 @@
 <template>
-    <RouterView v-if="isLogged" />
-    <LoginForm v-else />
+    <div>
+        <Sidenav v-if="isLogged" />
```
TatsuUkraine: this type of template breaks single responsibility, you need to define separate layout that will handle router view and rest parts and use it here or in route settings

frontend/src/components/layout/Sidepanel.vue
```diff
@@ -0,0 +1,109 @@
+<template>
```
TatsuUkraine: not quite get why you have 2 Sidepanels?

frontend/src/plugins/router.js
```diff
@@ -17,10 +17,6 @@ export default new Router({
     mode: 'history',
     base: '/',
     routes: [
-        {
```
TatsuUkraine: please return back this route, we need this to prevent empty page on root url

frontend/src/components/layout/header/UserControls.vue
```diff
@@ -0,0 +1,110 @@
...
+                    v-for="link in links"
+                    :key="link.text"
+                >
+                    <RouterLink
+                        :to="{ name: link.route }"
+                    >
+                        {{ link.text }}
+                    </RouterLink>
+                </VListItem>
+                <VListItem @click="logout">
```
TatsuUkraine: probably better create method for this event. We need not just logout user, but also navigate him to home screen

TatsuUkraine:
@wyyi also resolve please merge conflict

PR #66, opened 2019-08-10
fix user registration

naytis:


backend/app/Http/Controllers/Auth/RegisterController.php
```diff
@@ -5,7 +5,7 @@
 namespace App\Http\Controllers\Auth;
 
 use app\Actions\User\RegisterRequest;
-use app\Actions\User\RegisterUserAction;
+use app\Actions\User\RegisterAction;
```
TatsuUkraine: Ensure that you defined use in a right way, App should be with upper letter

backend/app/Actions/Auth/RegisterResponse.php
```diff
@@ -2,7 +2,7 @@
 
 declare(strict_types=1);
 
-namespace app\Actions\User;
+namespace app\Actions\Auth;
```
TatsuUkraine: Change app to App

PR #69, opened 2019-08-11
fixed login form styles

aborenko1996:


frontend/src/components/auth/LoginForm.vue
```diff
@@ -105,27 +105,12 @@
...
-                margin-top: 4px;
-            }
-        }
-
         &.v-input--is-focused {
-            border: 1px solid rgba(60, 87, 222, 0.52);
-            box-shadow: 0px 0px 14px rgba(194, 205, 223, 0.6);
+            ::v-deep {
+                .v-input__slot {
+                    border: 1px solid rgba(60, 87, 222, 0.52);
```
TatsuUkraine: you sure that you need this? vutify should already have border specified for inputs

aborenko1996: Without this property input will not have blue border on focus

TatsuUkraine: it seems that this going to be global styles, if so - it's better to define them in root component (where font definitions are placed)

TatsuUkraine:
@aborenko1996 lest remove border color from this component. We will have separate ticket to style inputs globally

PR #71, opened 2019-08-11
Add visitors page button component

LymarMaryna:


frontend/src/components/dashboard/visitors/ButtonComponent.vue
```diff
@@ -0,0 +1,95 @@
...
+                required: true
+            }
+        },
+        data () {
+            return {
+                iconSrs: ''
+            }
+        },
+        created () {
+            this.iconSrs = "data:image/png;base64," + getIcon(this.iconKey);
```
TatsuUkraine: it's better don't define images like that, let webpack do all optimization work

better create actual files in `assets/icons`, pass prop that will define icon name (key key that will be transpiled to icon name), this line better to remove

frontend/src/components/dashboard/visitors/ButtonComponent.vue
```diff
@@ -0,0 +1,95 @@
...
+                required: true
+            },
+            iconKey: {
+                type:String,
+                required: true
+            }
+        },
+        data () {
+            return {
+                iconSrs: ''
```
TatsuUkraine: this might no need

frontend/src/components/dashboard/visitors/ButtonComponent.vue
```diff
@@ -0,0 +1,95 @@
...
+        props: {
+            title: {
+                type:String,
+                required: true
+            },
+            character: {
+                type:String,
+                required: true
+            },
+            iconKey: {
```
TatsuUkraine: pass just icon name here

frontend/src/components/dashboard/visitors/ButtonComponent.vue
```diff
@@ -0,0 +1,95 @@
+<template>
+    <div
+        class="button-card bg-white d-inline-flex justify-content-start align-items-center"
+    >
+        <div
+            class="card-image"
+        >
+            <img
+                class="button-icon"
+                :src="iconSrs"
```
TatsuUkraine: to make it work with actual icon name change this row to
```suggestion
                :src="require(`@/assets/icons/${iconKey}.png`)"
```

frontend/src/components/dashboard/visitors/iconsConfig.js
```diff
@@ -0,0 +1,15 @@
+const icons = {
```
TatsuUkraine: create actually files for icons

PR #72, opened 2019-08-11
add endpoint for calculating bounce rate

naytis:
Tasks: [1](https://trello.com/c/f6ld55pU/90-visistors-add-endpoint-that-returns-calculated-bounce-rate) and [2](https://trello.com/c/EhLGmBCn/65-create-data-format-for-displaying-bounce-rate)


backend/app/Actions/Visitors/GetBounceRateAction.php
```diff
@@ -0,0 +1,27 @@
...
+    private $visitorRepository;
+
+    public function __construct(VisitorRepository $visitorRepository)
+    {
+        $this->visitorRepository = $visitorRepository;
+    }
+
+    public function execute(): GetBounceRateResponse
+    {
+        $visitors = $this->visitorRepository->withSinglePageInactiveSession()->count();
```
lenchvolodymyr: count is the aggregate function of a database.

And you should return data for a chart. It means it is not a single value, it should be a collection of bounce rates for a chosen date range.

lenchvolodymyr: You should create some GetBounceRateRequest that has some filters, like startDate and endDate.

lenchvolodymyr: And also some period: hour, day, week, month.

And according to these filters you should retrieve visitors that have only one finished session with one visit. And calculate bounce rate for each period

backend/app/Actions/Visitors/GetBounceRateAction.php
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    public function execute(GetBounceRateRequest $request): GetBounceRateResponse
+    {
+        $visitors = $this->visitorRepository
+            ->countSinglePageInactiveSessionBetweenDate(
+                $request->startDate(),
+                $request->endDate()
+            );
+        $allVisitors = $this->visitorRepository->countAll();
```
lenchvolodymyr: the rest of the visitors also should be retrieved in a date range

lenchvolodymyr:
Looks good. But can you write a test?

PR #73, opened 2019-08-11
Features/link add site with backend

KoR82:
All commits related with task/branch "Feature/add endpoint for adding website" have been reverted


frontend/src/store/modules/website/actions.js
```diff
@@ -8,38 +12,8 @@ export default {
...
-            } else if (fakeWebsiteData.domain !== newDataSite.domain) {
-                reject({
-                    errors: {
-                        domain: "Wrong domain of website"
-                    }
-                });
-            }
-
-        });
+        return requestService.create(resourceUrl, newDataSite)
```
TatsuUkraine: It's better to create API service that will responsible for updating data (see `api/auth.js` for example

TatsuUkraine:
It seems something went wrong, it can't find deps. Also ensure that you don't remove actions that are used in other components

frontend/src/store/modules/website/actions.js
```diff
@@ -8,40 +9,11 @@ export default {
...
-                    }
-                });
-            }
-            const fakeWebsiteData = {
-                name: 'Website name',
-                domain: 'http://domain.com',
-                single_page: newDataSite.single_page,
-                tracking_number: '123456789-10',
-            };
+        return addWebsite(newDataSite)
```
TatsuUkraine: Add check before data saving, that all needed data are ready to be saved, see prev mocked error throwing example for error types

frontend/src/store/modules/website/actions.js
```diff
@@ -16,30 +17,10 @@ export default {
...
-                });
-            } else if (fakeWebsiteData.domain !== newDataSite.domain) {
-                reject({
-                    errors: {
-                        domain: "Wrong domain of website"
-                    }
-                });
-            }
-
+            resolve();
```
TatsuUkraine: Best check I ever seen))))) I meant throw error with fields, that will force component, that emitted this action, navigate to specific page (see component implenation for more info), also you no need to wrap everything in promise, since your API method already returns promise, so you enough to just return it in action, but before you call this action, ensure that all data ready, reach me if you have any questions

frontend/src/store/modules/website/actions.js
```diff
@@ -1,47 +1,58 @@
...
-                        message: "Fill in all thi fields correctly."
-                    }
-                });
-            }
-            const fakeWebsiteData = {
-                name: 'Website name',
-                domain: 'http://domain.com',
-                single_page: newDataSite.single_page,
-                tracking_number: '123456789-10',
+        if(!newDataSite.name || !newDataSite.domain) {
```
TatsuUkraine: Please change this to more specific errors. For instance, if name not set, throw error with key name, if domain not set - error with key domain

TatsuUkraine:
@KoR82  please resolve conflicts and fix Code Quality issues

PR #76, opened 2019-08-12
Features/data format avg session

KoR82:


lenchvolodymyr:
Add endpoint that returns this time for chosen period

KoR82:
> Add endpoint that returns this time for chosen period

Done.

backend/app/Repositories/EloquentSessionRepository.php
```diff
@@ -6,11 +6,36 @@
...
 {
     public function getCollection(): Collection
     {
         return collect([]);
     }
+
+    public function getAvgSession(GetAvgSessionRequest $request): int
+    {
+
+        // it will be actual when User->Website relations is changed to OneToOne
```
lenchvolodymyr: get rid of comment

backend/app/Http/Resources/AvgSession.php
```diff
@@ -0,0 +1,24 @@
...
+class AvgSession extends JsonResource implements ApiResponse
+{
+    public function toArray($request): array
+    {
+        return [
+            'avg_session' => $this->convertSecondsToHoursMinutesSeconds($this->resource)
+        ];
+    }
+
+    private function convertSecondsToHoursMinutesSeconds(int $seconds): array
```
lenchvolodymyr: Better return timestamp in UTC, and UI will make a decision how to convert it to appropriate format

backend/app/Repositories/EloquentSessionRepository.php
```diff
@@ -6,11 +6,35 @@
...
     public function getCollection(): Collection
     {
         return collect([]);
     }
+
+    public function getAvgSession(GetAvgSessionRequest $request): int
+    {
+        $websiteId = auth()->user()->website->id;
+
+        $visitorsIDsOfWebsite = Visitor::where('website_id', $websiteId)
```
lenchvolodymyr: This is a problem. Do query visitors from VisitorRepository in GetAvgSessionAction. And moreover, Marina adds website_id to Session entity, so it should be redone in the future. But for now move this to action

backend/app/Repositories/EloquentSessionRepository.php
```diff
@@ -6,11 +6,35 @@
...
+                                        ->get()
+                                        ->pluck('id');
+
+        $avgSession = Session::whereIn('visitor_id', $visitorsIDsOfWebsite)
+                        ->where('start_session', '>=', Carbon::createFromTimestamp($request->startDate())->toDateString())
+                        ->where('start_session', '<=', Carbon::createFromTimestamp($request->endDate())->toDateString())
+                        ->get()
```
lenchvolodymyr: Here is responsibility of repository has finished. The rest calculation should be done inside Action

backend/app/Repositories/EloquentSessionRepository.php
```diff
@@ -6,11 +6,35 @@
...
+use App\Entities\Visitor;
 
 class EloquentSessionRepository implements SessionRepository
 {
     public function getCollection(): Collection
     {
         return collect([]);
     }
+
+    public function getAvgSession(GetAvgSessionRequest $request): int
```
lenchvolodymyr: instead of passing Request it is better to pass another DTO, for example AverageSessionFilter, with methods getVisitorsId(), getStartDate(), getEndDate(). And conversion to timestamp you should do outside of repository either in request or AverageSessionFilter

backend/tests/Feature/GetAvgSesiionTest.php
```diff
@@ -0,0 +1,69 @@
...
+        $user = factory(User::class)->create();
+        $website = factory(Website::class)->create();
+        $user->website()->save($website);
+        $token = JWTAuth::fromUser($user);
+        $websiteVisitors = factory(Visitor::class, 10)->create(['website_id' => $website->id]);
+        $start_session = Carbon::now();
+
+
+        $geoPositionId = factory(\App\Entities\GeoPosition::class)->create()->id;
+        $oSId = factory(\App\Entities\Os::class)->create()->id;
```
lenchvolodymyr: test should not be so complicated, why you cannot create session like this factory(Session::class, 10)->create() ?

backend/app/Actions/Sessions/GetAvgSessionResponse.php
```diff
@@ -0,0 +1,22 @@
...
+    private $avgSession;
+
+    public function __construct(Collection $avgSession)
+    {
+        $this->avgSession = $avgSession;
+    }
+
+    public function avgSession(): int
+    {
+        return (int) $this->avgSession->first()->avg;
```
lenchvolodymyr: DTO (Data Transfer Object) mustn't calculate anything, just retun a number. So calculation should be moved to the action

backend/app/Repositories/EloquentSessionRepository.php
```diff
@@ -6,11 +6,27 @@
...
 {
     public function getCollection(): Collection
     {
         return collect([]);
     }
+
+    public function getAvgSession(AverageSessionFilter $filter): Collection
+    {
+        return DB::table('sessions')
+                ->selectRaw('EXTRACT(EPOCH FROM (AVG(end_session - start_session))) as avg')
```
lenchvolodymyr: I cannot understand, if you're counting average here, what are you doing in the response then?

backend/tests/Feature/GetAvgSessionTest.php
```diff
@@ -0,0 +1,66 @@
...
+        $demographicId = factory(\App\Entities\Demographic::class)->create(['geo_position_id' => $geoPositionId])->id;
+        $deviceId = factory(\App\Entities\Device::class)->create()->id;
+        $systemId = factory(\App\Entities\System::class)->create([
+            'browser_id' => $browserId,
+            'os_id' => $oSId
+        ])->id;
+
+        factory(Session::class, 30)->create([
+            'visitor_id' => $websiteVisitors->pluck('id')->random(),
+            'start_session' => $start_session->addDays(rand(0, 4))->timestamp,
```
lenchvolodymyr: oh no, never use random in the test. The test must be repeatable, if the test will fail in some cases you will spend a lot of time to find why the test failing. The solution is don't use so many sessions. Create 4 or 5 sessions, some should be out of date range, some in range. And then compare with constant value result

backend/tests/Feature/GetAvgSessionTest.php
```diff
@@ -0,0 +1,66 @@
...
+            'device_id' => $deviceId ,
+            'system_id' => $systemId
+        ]);
+
+        $queryString = '?filter[startDate]='.Carbon::now()->timestamp.'&filter[endDate]='.Carbon::now()->addDays(5)->timestamp;
+
+        $this->actingAs($user)
+            ->withHeader('Authorization', "Bearer $token")
+            ->getJson('/api/v1/sessions/average/'.$queryString)
+            // ->assertOk()
```
lenchvolodymyr: uncomment this

backend/tests/Feature/GetAvgSessionTest.php
```diff
@@ -0,0 +1,66 @@
...
+            'entrance_page_id' => $pageId,
+            'demographic_id' => $demographicId ,
+            'device_id' => $deviceId ,
+            'system_id' => $systemId
+        ]);
+
+        $queryString = '?filter[startDate]='.Carbon::now()->timestamp.'&filter[endDate]='.Carbon::now()->addDays(5)->timestamp;
+
+        $this->actingAs($user)
+            ->withHeader('Authorization', "Bearer $token")
```
lenchvolodymyr: actually actingAs is enough

backend/tests/Feature/GetAvgSessionTest.php
lenchvolodymyr: And pay attention, that some of the sessions shouldn't belong to website of user

backend/tests/Feature/GetAvgSessionTest.php
```diff
@@ -0,0 +1,66 @@
...
+            ->getJson('/api/v1/sessions/average/'.$queryString)
+            // ->assertOk()
+            ->assertJsonStructure([
```
lenchvolodymyr: for this task you should check exactly value that resource returns

backend/tests/Feature/GetAvgSesiionTest.php
KoR82: In this case test can not be passed because SessionFactory requires related entities (Visitor, Page etc.) to be created before seeding sessions table.
I want to check in this test successful api response with real sql query.

backend/app/Repositories/EloquentSessionRepository.php
KoR82: Here i get integer, for ex. 100 sec. You asked response has to be in timestamp UTC, therefore I convert seconds to timestamp in the response and get: ['avg_sessions => '1970-01-01 00:01:40'']

backend/tests/Feature/GetAvgSessionTest.php
```diff
@@ -0,0 +1,107 @@
...
+use Illuminate\Foundation\Testing\WithFaker;
+use Illuminate\Foundation\Testing\RefreshDatabase;
+use App\Entities\User;
+use App\Entities\Website;
+use App\Entities\Visitor;
+use App\Entities\Session;
+use App\Entities\Page;
+use App\Entities\System;
+use App\Entities\GeoPosition;
+use App\Entities\Demographic;
```
lenchvolodymyr: this entity does not exist anymore

backend/tests/Feature/GetAvgSessionTest.php
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    private function seedDataBase(): void
+    {
+        factory(Website::class)->create();
+        factory(Visitor::class)->create();
+        factory(Page::class)->create();
+        factory(System::class)->create();
+        factory(GeoPosition::class)->create();
+        factory(Demographic::class)->create();
```
lenchvolodymyr: Demographic was deleted

backend/app/Actions/Sessions/GetAvgSessionRequest.php
```diff
@@ -0,0 +1,37 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Actions\Sessions;
+
+use App\Http\Requests\Api\GetAvgSessionHttpRequest;
+
+final class GetAvgSessionRequest
```
lenchvolodymyr: inherit from ButtonDataRequest. There is already exists method period, that covers all cases for start and end dates. This is also good, but just to be consistent with the whole project

PR #78, opened 2019-08-12
Add website adding master guard

LymarMaryna:


frontend/src/plugins/router.js
```diff
@@ -69,6 +71,7 @@ export default new Router({
                 {
                     path: 'websites/add',
                     component: AddWebsitePage,
+                    redirect: { name: 'add_websites_step_1' },
```
TatsuUkraine: no quite sure if this works, if you need redirect to from default route to some internal route, use child with empty path (default path)

frontend/src/plugins/router.js
```diff
@@ -77,7 +80,15 @@ export default new Router({
                             meta: {
                                 step: 1
                             },
-                            props: true
+                            props: true,
+                            beforeEnter: (to, from, next) => {
+                                if (store.state.isCurrentWebsite) {
```
TatsuUkraine: this probably won't work like it should be, since route hook can be called before data will be fetched

frontend/src/plugins/router.js
```diff
@@ -86,6 +97,14 @@ export default new Router({
                             meta: {
                                 step: 2
                             },
+                            beforeEnter: (to, from, next) => {
+                                if (store.state.isCurrentWebsite) {
```
TatsuUkraine: same as in prev route hook

frontend/src/store/modules/auth/actions.js
```diff
@@ -13,6 +15,12 @@ export default {
                 const user = response.data;
                 context.commit(SET_AUTHENTICATED_USER, user);
 
+                getCurrentUserWebsite()
```
TatsuUkraine: it's probably not the best place to fetch data, since you mixin contexts here

PR #81, opened 2019-08-12
Feature/line chart component

TumkoVolodymyr:


frontend/src/components/common/LineChart.vue
```diff
@@ -0,0 +1,158 @@
...
+        },
+
+        computed: {
+            chartData(){
+                let tooltipObj = {'type': 'string', 'role': 'tooltip', 'p': {'html': true}};
+                let pointStyle = 'point { stroke-color: #3C57DE; size: 5.5; shape-type: circle; fill-color: #FFFFFF; }';
+                let data = [];
+                data.push([{type: 'string', name: 'xLabel'}, '', 'yValue', {'type': 'string', 'role': 'style'}, tooltipObj]);
+
+                this.data.forEach( (element) => {
```
lenchvolodymyr: If you need to return updated array better use map, instead of data and forEach

frontend/src/components/common/LineChart.vue
```diff
@@ -0,0 +1,158 @@
...
+            },
+
+        }
+    }
+</script>
+
+<style lang="scss">
+    div.google-visualization-tooltip {
+        margin-left: -100px !important;
+        z-index:+1;
```
lenchvolodymyr: +1 is it okay?

PR #82, opened 2019-08-12
Feature/add auth by update page

SerhiiPlotnikov:


frontend/src/components/auth/Spinner.vue
```diff
@@ -0,0 +1,10 @@
+<template>
+    <div class="text-center">
+        <VOverlay value="true">
```
TatsuUkraine: use `:` if you want to pass boolean

PR #85, opened 2019-08-13
fix user name render in topnavbar, make signin button into router link

wyyi:


frontend/src/components/layout/header/UserControls.vue
```diff
@@ -14,7 +14,7 @@
             </svg>
         </VBtn>
         <VToolbarTitle class="hidden-sm-and-down username mr-6">
-            Hello, <span>{{ user.name }}</span>
+            Hello, <span>{{ user().name }}</span>
```
TatsuUkraine: No need to do like this, original issue is that mapGetter should be placed in `computed` attr, and not `methods` attr inside component definition

PR #98, opened 2019-08-15
Add endpoint to get visitors by params

LymarMaryna:


lenchvolodymyr:
Looks good. But could you write some tests, just to be sure that logic works well? And also fix style issues and some tests were failed. Make sure that those tests are still actual.

PR #104, opened 2019-08-15
Fix logo redirect

SerhiiPlotnikov:


frontend/src/components/layout/header/TopNavbar.vue
```diff
@@ -7,7 +7,7 @@
     >
         <VToolbarTitle class="pl-4">
             <span class="logo pr-3">
-                <RouterLink :to="{ name: 'home' }">M</RouterLink>
+                <RouterLink :to="{ name: isLoggedIn?'dashboard':'home' }">M</RouterLink>
```
TatsuUkraine: define computed prop, that will return route object `{ name:.....}`

PR #108, opened 2019-08-15
mock router push error catch to empty function in global context

TumkoVolodymyr:
https://trello.com/c/WjSspQZm


TatsuUkraine:
lets wait until routing will settle down, just to avoid any unpredictable behavior for other devs

PR #125, opened 2019-08-15
Feature/singin page design

TumkoVolodymyr:
https://trello.com/c/3zxn35Sq


frontend/src/components/auth/LoginForm.vue
```diff
@@ -1,62 +1,72 @@
 <template>
-    <VContent>
-        <VFlex
-            lg6
-            md6
-            sm12
-            xs12
-            :class="{'mx-5': $vuetify.breakpoint.smAndUp}"
+    <div class="form">
+        <Spinner
```
lenchvolodymyr: On the meeting, we came up with a decision not to block a whole form, but only a button

frontend/src/components/auth/LoginForm.vue
```diff
@@ -101,23 +109,55 @@
             }),
             onLogin() {
                 if (this.$refs.form.validate()) {
+                    this.isLoading = true;
                     this.login({
                         email: this.email,
                         password: this.password
                     }).then(() => {
+                        this.isLoading = false;
```
TatsuUkraine: no need to duplicate same action twice, just add `then` after catch statement or `finally`, and put this prop change there

frontend/src/components/auth/LoginForm.vue
```diff
@@ -101,23 +109,55 @@
...
                     }).catch((error) => {
+                        this.isLoading = false;
                         this.showErrorMessage(error);
                     });
                 }
             }
+        },
+        computed: {
+            signInText() {
+                return this.isLoading?'SIGN INNING...':'SIGN IN';
```
TatsuUkraine: not sure that such word like "SIGN INNING" exists, better replace it to "Processing..." or something like that

PR #131, opened 2019-08-16
fix: edit user page

mes1901:
https://trello.com/c/T0wKGeyY

### DIFF
--- a/frontend/src/components/auth/UpdateProfileForm.vue
+++ b/frontend/src/components/auth/UpdateProfileForm.vue
@@ -20,7 +20,6 @@
                     v-model="editUser.name"
                     type="text"
                     :rules="nameRules"
-                    required
                     solo
                 />
                 <label class="caption grey--text mt-2 mb-3">
@@ -32,7 +31,6 @@
                     v-model="editUser.email"
                     type="email"
                     :rules="emailRules"
-                    required
                     solo
                 />
                 <label class="caption grey--text mt-2 mb-3">
@@ -46,7 +44,6 @@
                     :type="passwordType"
                     counter
                     :rules="passwordRules"
-                    required
                     solo
                     @click:append="showPassword = !showPassword"
                 />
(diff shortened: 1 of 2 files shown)


frontend/src/components/auth/UpdateProfileForm.vue
```diff
@@ -97,19 +93,16 @@
...
+                    v => validateFullName(v) || 'Enter the correct information'
                 ],
                 emailRules: [
-                    v => !!v || 'E-mail is required',
                     v => validateEmail(v) || 'E-mail must be valid',
                 ],
                 passwordRules: [
-                    v => !!v || 'Password is required',
-                    v => validatePassword(v) || 'Password must be equal or more than 8 characters'
+                    v => (v.length === 0 || v.length >= 8) || 'Password must be equal or more than 8 characters'
```
lenchvolodymyr: why did not you change validatePassword method instead?

frontend/src/services/validation.js
```diff
@@ -1,4 +1,5 @@
 const emailRegex = /^(([^<>()[\]\\.,;:\s@"]+(\.[^<>()[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/;
+const fullNameRegex = /^[a-z]([-']?[a-z]+)*( [a-z]([-']?[a-z]+)*)+$/i;
```
lenchvolodymyr: why did you add so complicated reg exp for user name? This won't pass cyrillic symbols, and any instead of latin.

For full name only one validation rule should be accepted: not empty

frontend/src/components/auth/UpdateProfileForm.vue
lenchvolodymyr: I understood. But better you do v.length === 0 || validatePassword(v)

PR #146, opened 2019-08-18
Fix/spa should match with previously set

SerhiiPlotnikov:
Update action doesn't work correctly, or maybe this action hasn't implemented yet completely. It change  website name, and reset SPA switcher to false position

### DIFF
--- a/frontend/src/components/website/WebsiteForm.vue
+++ b/frontend/src/components/website/WebsiteForm.vue
@@ -39,7 +39,7 @@
                 <div class="inline-element mb-7">
                     <label>SPA</label>
                     <VSwitch
-                        :value="currentWebsite.single_page"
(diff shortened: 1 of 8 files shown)


frontend/src/components/website/WebsiteForm.vue
```diff
@@ -39,7 +39,7 @@
                 <div class="inline-element mb-7">
                     <label>SPA</label>
                     <VSwitch
-                        :value="currentWebsite.single_page"
+                        v-model="currentWebsite.single_page"
```
TatsuUkraine: Never mutate state outside of mutation. Possibly there 2 types of issue happening - or Switch component doesn't have prop with name value, or backend doesn't return right value. In any case state shouldn't be passed to v-model directive directly, since it can cause state mutation outside of store

PR #165, opened 2019-08-19
Feature/add endpoint for count of visitors

SerhiiPlotnikov:


backend/routes/api.php
```diff
@@ -78,6 +78,13 @@
         ], function () {
             Route::get('/', 'VisitorController@getNewVisitorsByDateRange');
         });
+
+
+        Route::group([
+            'prefix'=>'button-visitors'
+        ], function () {
+            Route::get('/count', 'VisitorController@getVisitorsCount');
```
lenchvolodymyr: rename router just for /button-visitors/

backend/tests/Feature/ButtonVisitorsApiTest.php
```diff
@@ -0,0 +1,74 @@
...
+    protected function setUp(): void
+    {
+        parent::setUp();
+        $this->user = factory(User::class)->create();
+    }
+
+    public function testButtonCountVisitors()
+    {
+        factory(Website::class)->create();
+        $firstDate = new DateTime('@1565560845');
```
lenchvolodymyr: use instead readable dates, and transform to timestamps where it is necessary. It will be easier to support

backend/app/Repositories/EloquentButtonVisitorsRepository.php
```diff
@@ -0,0 +1,20 @@
...
+
+
+namespace App\Repositories;
+
+use App\Entities\Visitor;
+use App\Repositories\Contracts\ButtonVisitorsRepository;
+
+class EloquentButtonVisitorsRepository implements ButtonVisitorsRepository
+{
+    public function getVisitorsCount(string $startData, string $endData, int $websiteId, int $userId): int
```
lenchvolodymyr: why don't you pass DatePeriod instead of $startData and $endData? And data means different than date.

PR #168, opened 2019-08-19
fix/style-css

yashuk803:


yashuk803:
https://trello.com/c/BaBqqcCc/156-add-white-background-for-the-half-of-page-master-of-adding-website
https://trello.com/c/6ejEcpId/159-add-paddings-to-the-main-layout-according-to-design

frontend/src/scss/main.scss
```diff
@@ -56,4 +56,17 @@ svg path {
 
 .v-btn {
     text-decoration: none !important;
-}
\ No newline at end of file
+}
+
+.content-with-padding {
+    padding: 70px 66px 0 80px;
```
lenchvolodymyr: what are the magic numbers?) And why is bottom without padding?

lenchvolodymyr: There is really small distance from header to content.

lenchvolodymyr: And these styles are probably better

width: 90%;
    margin: 0 auto;
    padding: 32px 0;

PR #169, opened 2019-08-19
fix login form component syle

TumkoVolodymyr:
https://trello.com/c/IEtQJunm


TatsuUkraine:
it seems we going to have different approach in login guard here #179

PR #174, opened 2019-08-19
add period service on frontend

TumkoVolodymyr:
https://trello.com/c/qA2Xok98

### DIFF
--- a/frontend/src/services/periodService.js
+++ b/frontend/src/services/periodService.js
@@ -1,0 +1,77 @@
+const PERIOD_TODAY = 'today';
+const PERIOD_YESTERDAY = 'yesterday';
+const PERIOD_LAST_WEEK = 'last_week';
+const PERIOD_LAST_MONTH = 'last_month';
+const PERIOD_LAST_QUARTAL = 'last_quartal';
+const PERIOD_ALL = 'all_period';
+
+export const DAYS_IN_WEEK = 7;
+export const HOURS_IN_DAY = 24;
+export const MINUTES_IN_HOUR = 60;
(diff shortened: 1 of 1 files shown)


frontend/src/services/periodService.js
```diff
@@ -0,0 +1,77 @@
+const PERIOD_TODAY = 'today';
+const PERIOD_YESTERDAY = 'yesterday';
+const PERIOD_LAST_WEEK = 'last_week';
+const PERIOD_LAST_MONTH = 'last_month';
+const PERIOD_LAST_QUARTAL = 'last_quartal';
+const PERIOD_ALL = 'all_period';
+
+export const DAYS_IN_WEEK = 7;
```
TatsuUkraine: not quite sure that you need to export all this con values

TatsuUkraine:
@TumkoVolodymyr issue still remains - all this service can be simplified a lot if you use momentjs, if not - let me know

TumkoVolodymyr:
> @TumkoVolodymyr issue still remains - all this service can be simplified a lot if you use momentjs, if not - let me know

@TatsuUkraine  can we left as it is for this time?

TatsuUkraine:
> > @TumkoVolodymyr issue still remains - all this service can be simplified a lot if you use momentjs, if not - let me know
> 
> @TatsuUkraine can we left as it is for this time?

@TumkoVolodymyr for what time?) this will never be changed if commit it like this) try to make appropriate implementation without reinventing a wheel. If you get any problems - let me know. I would also recommend to implement period service in separate branch, so if you stuck with it and everyone else will be block - we could merge this variant

PR #176, opened 2019-08-20
Feature/add period dropdown on the page visitors

SerhiiPlotnikov:


frontend/src/components/auth/LoginForm.vue
```diff
@@ -139,7 +139,7 @@
         .v-btn {
             text-transform: none;
 
-            +.start {
+            + .start {
```
lenchvolodymyr: ??

frontend/src/store/modules/visitors/state.js
```diff
@@ -0,0 +1,3 @@
+export default {
+    selectedPeriod: undefined
```
lenchvolodymyr: default: last_week

frontend/src/components/dashboard/visitors/PeriodDropdown.vue
```diff
@@ -0,0 +1,87 @@
...
+                    {title: 'Last 7 days', value: 'last_week'},
+                    {title: 'Last month', value: 'last_month'},
+                    {title: 'Last quartal', value: 'last_quartal'},
+                    {title: 'All period', value: 'all_period'}
+                ]
+            };
+        },
+        methods: {
+            ...mapActions('visitors', {
+                changeSelectedPeriod: CHANGE_SELECTED_PERIOD
```
TatsuUkraine: not quite sure that you need Vuex store for this component, in general Vuex store shouldn't be used everywhere and in any case. In some cases consider to use local state with common components

For instance, it does make sense to specify selected value as prop and use v-model to track value change in parent component. Also take in account the if you using Vuex for every state management you Vuex store quite quickly become unsupported massive object. And also - Vuex will persist data, so each time when user will get back to this particular component this state will be restored. In addition to that - using Vuex for every key means that you will need to track and clear a lot of stuff on logout

frontend/src/components/dashboard/visitors/PeriodDropdown.vue
```diff
@@ -0,0 +1,87 @@
...
+    >
+        <VSelect
+            :items="items"
+            item-text="title"
+            item-value="value"
+            class="option-select"
+            v-model="selectedItem"
+            solo
+            return-object
+            @change="change"
```
TatsuUkraine: this might be redundant, since you already using v-model on this component

PR #179, opened 2019-08-20
Changing the logic of guards

LymarMaryna:
https://trello.com/c/ozCbEiYZ
https://trello.com/c/iJH1VzUN


frontend/src/components/layout/header/UserControls.vue
```diff
@@ -74,8 +75,12 @@
             ...mapActions('auth', {
                 logout: LOGOUT
             }),
+            ...mapActions('website', {
+                resetData: RESET_DATA
+            }),
             endSession() {
                 this.logout();
+                this.resetData();
```
TatsuUkraine: it's better to dispatch this action in LOGOUT action, if it creates any issues, keep it like it now, but if not - better to keep LOGOUT action centralized since logout can happen not only here

PR #193, opened 2019-08-21
Feature/add period service

yashuk803:
https://trello.com/c/qA2Xok98/173-add-periodservice


frontend/src/services/periodService.js
```diff
@@ -0,0 +1,81 @@
...
+            endDate = currentDate();
+            break;
+        case 'last_quartal':
+            interval = getDayInterval(30);
+            startDate = getSubtractStartMonth(3);
+            endDate = currentDate();
+            break;
+        case 'all_period':
+            interval = getDayInterval(30);
+            startDate = NaN;
```
TatsuUkraine: I would just use undefined in that case

frontend/src/services/periodService.js
```diff
@@ -0,0 +1,81 @@
...
+            endDate = currentDate();
+            break;
+        default:
+            throw 'This period is not defined';
+    }
+
+    return {
+        interval: interval,
```
TatsuUkraine: Use better to use shortcuts like
```
return {
    interval,
    startDate,
    endDate
}
```
In that way if any of those variables will be undefined, result object won't contain key with it's name. Just a recommendation, it doesn't block PR from approval

frontend/src/services/periodService.js
```diff
@@ -0,0 +1,81 @@
...
+
+const getDayInterval = (day) => {
+    return moment.duration({d: day}).asSeconds();
+};
+
+const getWeekInterval = (week) => {
+    return moment.duration({w: week}).asSeconds();
+};
+
+const getTimeByPeriod = (period) => {
```
TatsuUkraine: You can just export this function, and remove default export

PR #196, opened 2019-08-21
fix submenu is opened for websiteinfo

KoR82:


frontend/src/components/layout/Sidepanel.vue
```diff
@@ -102,7 +104,36 @@
                     ]
                 },
             ]
-        })
+        }),
+
+        watch:{
+            $route (to, from) {
+                console.log(this.$route.name, to, from);
```
lenchvolodymyr: remove console.log

PR #203, opened 2019-08-22
add-page-views-to-audience

mes1901:
https://trello.com/c/EChDghOS


frontend/src/components/dashboard/page_views/ButtonComponent.vue
```diff
@@ -0,0 +1,126 @@
+<template>
```
TatsuUkraine: don't copy/paste, create 1 common component, that will work with just props, use it here and in visitors page

frontend/src/components/dashboard/page_views/PeriodDropdown.vue
```diff
@@ -0,0 +1,91 @@
+<template>
```
TatsuUkraine: don't copy/paste - make 1 common component that will be working with just props, and use it here and in Visitors page

frontend/src/pages/PageViews.vue
```diff
@@ -0,0 +1,186 @@
...
+    import {isWebsite} from '../mixins/isWebsite';
+    import {
+        PAGE_VIEWS,
+        UNIQUE_PAGE_VIEWS,
+        ACTIVE_USERS,
+        BOUNCE_RATE
+    } from '../configs/page_views/buttonTypes.js';
+
+    export default {
+        mixins: [isWebsite],
```
TatsuUkraine: don't put guard on each child page, you have 1 page that wraps all of them, which means that guard mixin should be used only there (it also means that Visitors page no longer need it too)

TatsuUkraine:
@mes1901 reach me you have any questions

frontend/src/components/dashboard/common/PeriodDropdown.vue
```diff
@@ -9,44 +9,38 @@
...
+        },
         data() {
             return {
                 items: [
                     {title: 'Today', value: 'today'},
                     {title: 'Yesterday', value: 'yesterday'},
                     {title: 'Last 7 days', value: 'last_week'},
-                    {title: 'Last month', value: 'last_month'},
-                    {title: 'Last quartal', value: 'last_quartal'},
+                    {title: 'Last 30', value: 'last_month'},
```
lenchvolodymyr: it was okay

PR #207, opened 2019-08-22
add enpoint for creation visits

naytis:
https://trello.com/c/GaKjgPX0/199-add-endpoint-for-creation-visits


backend/app/Actions/Visits/CreateVisitAction.php
```diff
@@ -0,0 +1,181 @@
...
+        $geoPosition = $this->getOrCreateGeoPosition($request->ip());
+
+        $session = $this->getOrCreateSession(
+            $visitor->id,
+            $page->id,
+            $request->language(),
+            $system->id
+        );
+
+        $visit = Visit::create([
```
lenchvolodymyr: entity cannot set itself. It is problem of eloquent and active record pattern at all. It should be saved by corresponding repository

backend/app/Repositories/EloquentSessionRepository.php
```diff
@@ -38,4 +36,10 @@ public function getAvgSession(AverageSessionFilter $filter): Collection
                 ->where('start_session', '<=', $filter->getEndDate())
                 ->get();
     }
+
+    public function lastActiveByVisitorId(int $visitorId): ?Session
+    {
+        return Session::whereVisitorId($visitorId)
+            ->whereTime('updated_at', '>', (Carbon::now())->subMinutes(30)->toDateTimeString())->first();
```
lenchvolodymyr: if I am not mistaken, visitor has particular property instead of updated_at: last_activity

backend/app/Actions/Visits/CreateVisitAction.php
```diff
@@ -0,0 +1,181 @@
...
+        $this->geoPositionRepository = $geoPositionRepository;
+    }
+
+    public function execute(CreateVisitRequest $request): CreateVisitResponse
+    {
+        $token = Str::after($request->token(), 'Bearer ');
+        JWTAuth::setToken($token);
+        $visitorId = JWTAuth::getPayload()->get('visitor_id');
+
+        $visitor = $this->visitorRepository->getById($visitorId);
```
lenchvolodymyr: also last_activity should be updated

backend/app/Actions/Visits/CreateVisitAction.php
```diff
@@ -0,0 +1,181 @@
...
+            $agent->platform(),
+            $request->device(),
+            $agent->browser(),
+            $request->resolutionHeight(),
+            $request->resolutionWidth()
+        );
+
+        $geoPosition = $this->getOrCreateGeoPosition($request->ip());
+
+        $session = $this->getOrCreateSession(
```
lenchvolodymyr: end_session property also should be updated

lenchvolodymyr: end_session means last_activity_session, but it is still used with this meaning

backend/routes/api.php
```diff
@@ -61,6 +61,7 @@
             'prefix' => 'visits'
         ], function () {
             Route::get('/by-table', 'VisitController@getPageViewsByParameter');
+            Route::post('/', 'VisitController@createVisit');
```
lenchvolodymyr: it is external api, so there will not be authenticated users

backend/app/Http/Controllers/Api/VisitController.php
```diff
@@ -58,4 +62,11 @@ public function getPageViewsCountForFilterData(GetPageViewsCountFilterHttpReques
         $response = $this->getPageViewsCountAction->execute(GetPageViewsCountRequest::fromRequest($request));
         return ApiResponse::success(new ButtonResource($response));
     }
+
+    public function createVisit(CreateVisitHttpRequest $request): ApiResponse
+    {
+        $response = $this->createVisitAction->execute(CreateVisitRequest::fromRequest($request));
+
+        return ApiResponse::success(new VisitResource($response->visit()));
```
lenchvolodymyr: as it is not internal Api, it shouldn't return any data in response, status 200 is enough. And probably it should be in another namespace. Like OpenApi, or something

backend/app/Repositories/EloquentSessionRepository.php
lenchvolodymyr: and still, visitor has particular property last_activity

backend/tests/Feature/VisitsApiTest.php
```diff
@@ -198,4 +204,39 @@ public function testPageViewsFailedInterval()
...
+            'REMOTE_ADDR' => $ip,
+            'X-Visitor' => 'Bearer ' . $token
+        ];
+
+        $url = 'api/v1/visits/';
+
+        $this->actingAs($this->user)
+            ->json('POST', $url, $data, $headers)
+            ->assertStatus(200)
+            ->assertJsonStructure([]);
```
lenchvolodymyr: and check that database has particular records

backend/app/Repositories/EloquentSessionRepository.php
lenchvolodymyr: Okay, I see, it should be end_session so, as it the same as last_activity of session. Or date of last session visit

backend/app/Actions/Visits/CreateVisitAction.php
```diff
@@ -0,0 +1,181 @@
...
+
+        if ($session !== null) {
+            $this->sessionRepository->updateEndSession($session);
+            return $session;
+        }
+
+        $session = new Session();
+
+        $session->start_session = Carbon::now();
+        $session->visitor_id = $visitorId;
```
lenchvolodymyr: end_session also should be set, as it points on last activity of session. Queries won't work for first session with one visit

PR #206, opened 2019-08-22
add action for retrieving buttonData

mes1901:
https://trello.com/c/V900lC2R

PR #213, opened 2019-08-22
Feature/add action for retrieving table data

StasKozar:
https://trello.com/c/f9VWMjCQ/210-add-action-for-retrieving-tabledata


TatsuUkraine:
It seems you branch is extended from https://github.com/BinaryStudioAcademy/bsa-2019-metrica/pull/206 which means that you PR contains exact same issues, don't do this anymore, since you PR will be blocked until #206 will be merged

frontend/src/store/modules/visitors/actions.js
```diff
@@ -16,4 +33,30 @@ export default {
             context.commit(RESET_BUTTON_FETCHING, data.button);
         }
     },
+    [FETCH_TABLE_DATA]: (context, data) => {
+        if (!data.value) {
+            return;
+        }
+        context.commit(SET_TABLE_DATA_FETCHING);
+        context.commit(GET_ACTIVE_BUTTON)
```
lenchvolodymyr: you will not get data wit this approach. Take active button from context.state

frontend/src/store/modules/visitors/actions.js
```diff
@@ -16,4 +33,30 @@ export default {
...
+            })
+            .then(response => {
+                context.commit(SET_TABLE_DATA, response.data);
+            })
+            .catch(err => {
+                context.commit(RESET_TABLE_DATA_FETCHING);
+                throw err;
+            })
+            .finally(() => {
+                context.commit(RESET_TABLE_DATA_FETCHING);
```
lenchvolodymyr: probably better to move it in then, as if error will occur the error can be lost and unhandled

frontend/src/store/modules/visitors/actions.js
StasKozar: Fixed

frontend/src/store/modules/visitors/actions.js
StasKozar: Moved

PR #218, opened 2019-08-22
Feature/add action for retrieving chartdata

devhackerone:


frontend/src/store/modules/visitors/actions.js
```diff
@@ -30,6 +31,29 @@ export default {
             context.commit(RESET_BUTTON_FETCHING, data.button);
         }
     },
+    [GET_LINE_CHART_DATA]: (context, data) => {
+        if (!data.value) {
+            return;
+        }
+
+        context.commit(SET_LINE_CHART_DATA_FETCHING);
+        context.commit(GET_SELECTED_PERIOD)
```
TatsuUkraine: commit doesn't return promise, also mutation shouldn't return data, use state from context object

in general all chain, that implemented below, won't work

frontend/src/store/modules/visitors/actions.js
```diff
@@ -30,6 +31,29 @@ export default {
             context.commit(RESET_BUTTON_FETCHING, data.button);
         }
     },
+    [GET_LINE_CHART_DATA]: (context, data) => {
```
TatsuUkraine: don't call action as getter, unless it returns anything, in this particular case - it doesn't and shouldn't return anything

frontend/src/api/visitors/averageSessionsService.js
```diff
@@ -60,3 +60,5 @@ export const averageSessionService = {
     fetchChartValues,
     fetchTableValues
 };
+
+export default averageSessionService;
```
TatsuUkraine: if you going to export this kind of default, you no need to export const above

frontend/src/api/visitors/bounceRateService.js
```diff
@@ -60,3 +60,5 @@ export const bounceRateService = {
     fetchChartValues,
     fetchTableValues
 };
+
+export default bounceRateService;
```
TatsuUkraine: if you going to export this kind of default, you no need to export const above

frontend/src/api/visitors/newVisitorsService.js
```diff
@@ -60,3 +60,5 @@ export const newVisitorsService = {
     fetchChartValues,
     fetchTableValues
 };
+
+export default newVisitorsService;
```
TatsuUkraine: if you going to export this kind of default, you no need to export const above

frontend/src/api/visitors/sessionsService.js
```diff
@@ -60,3 +60,5 @@ export const sessionsService = {
     fetchChartValues,
     fetchTableValues
 };
+
+export default sessionsService;
```
TatsuUkraine: if you going to export this kind of default, you no need to export const above

frontend/src/api/visitors/totalVisitorsService.js
```diff
@@ -60,3 +60,5 @@ export const totalVisitorsService = {
     fetchChartValues,
     fetchTableValues
 };
+
+export default totalVisitorsService;
```
TatsuUkraine: if you going to export this kind of default, you no need to export const above

frontend/src/api/visitors/visitsService.js
```diff
@@ -60,3 +60,5 @@ export const visitsService = {
     fetchChartValues,
     fetchTableValues
 };
+
+export default visitsService;
```
TatsuUkraine: if you going to export this kind of default, you no need to export const above

frontend/src/pages/Visitors.vue
```diff
@@ -71,7 +71,7 @@
     import {mapGetters} from 'vuex';
     import {
         GET_PIE_CHART_DATA,
-        GET_LINE_CHART_DATA,
+        FETCH_LINE_CHART_DATA,
```
TatsuUkraine: getter should fetch anything

frontend/src/store/modules/visitors/types/getters.js
```diff
@@ -5,5 +5,7 @@ export const GET_TABLE_DATA_FETCHING = 'getTableDataFetching';
 export const GET_BUTTON_DATA = 'getButtonData';
 export const GET_ACTIVE_BUTTON = 'getActiveButton';
 export const GET_PIE_CHART_DATA = 'getPieChartData';
-export const GET_LINE_CHART_DATA = 'getChartData';
-export const FETCH_TABLE_DATA = 'getTableData';
+export const GET_LINE_CHART_ITEMS = 'getLineChartItems';
+export const GET_LINE_CHART_FETCHING = 'getLineChartFetching';
+export const FETCH_LINE_CHART_DATA = 'getChartData';
```
TatsuUkraine: getters will never fetch anything so getter  name, that starts with `FETCH_` doesn't make sense

frontend/src/api/visitors/averageSessionsService.js
devhackerone: removed all export default ...Service;

frontend/src/pages/Visitors.vue
devhackerone: back GET_LINE_CHART_DATA

frontend/src/store/modules/visitors/actions.js
```diff
@@ -40,6 +46,34 @@ export default {
...
+            context.commit(RESET_LINE_CHART_FETCHING);
+        }
+    },
+    [FETCH_LINE_CHART_DATA]: (context, data) => {
+        if (!data.value) {
+            return;
+        }
+        context.commit(SET_LINE_CHART_FETCHING);
+
+        getTimeByPeriod(context.state.selectedPeriod)
```
TatsuUkraine: again, `getTimeByPeriod` doesn't return Promise, it returns object. So it won't have `then` method, so this action won't work

frontend/src/store/modules/visitors/actions.js
```diff
@@ -57,7 +91,7 @@ export default {
         }
         context.commit(SET_TABLE_FETCHING);
 
-        getTimeByPeriod(context.state.selectedPeriod)
+        return getTimeByPeriod(context.state.selectedPeriod)
```
TatsuUkraine: again, `getTimeByPeriod` doesn't return Promise, it returns object. So it won't have `then` method, so this action won't work

frontend/src/store/modules/visitors/actions.js
```diff
@@ -40,6 +46,36 @@ export default {
...
+    },
+    [FETCH_LINE_CHART_DATA]: (context, data) => {
+        if (!data.value) {
+            return;
+        }
+        context.commit(SET_LINE_CHART_FETCHING);
+        const period = getTimeByPeriod(context.state.selectedPeriod);
+        const startDate = period.startDate;
+        const endDate = period.endDate;
+        let chartData = [];
```
TatsuUkraine: this variable is redundant

frontend/src/store/modules/visitors/actions.js
```diff
@@ -40,6 +46,36 @@ export default {
...
+
+        return factoryVisitorsService.create(context.state.activeButton)
+            .fetchChartValues(startDate.unix(), endDate.unix(), data.groupedParameter)
+                .then(response => {
+                    chartData.items = response.data;
```
TatsuUkraine: just for future - Array and Object it's to different stuff, if you want to set attributes in variable, don't define it as array

PR #223, opened 2019-08-22
Fix/change response of bounce rate button

SerhiiPlotnikov:


backend/app/Actions/Visitors/GetBounceRateAction.php
```diff
@@ -27,11 +28,11 @@ public function execute(GetBounceRateRequest $request): GetBounceRateResponse
             );
 
         if ($allVisitors === 0) {
-            return new GetBounceRateResponse(0);
+            return new ButtonValue((string)0);
```
lenchvolodymyr: (string)0 === ""

PR #238, opened 2019-08-23
Feature/add email confirmation after registration

SerhiiPlotnikov:


backend/app/Entities/User.php
```diff
@@ -40,6 +41,11 @@ public function sendPasswordResetNotification($token)
         $this->notify(new MailResetPasswordNotification($token));
     }
 
+    public function sendSuccessRegistrationNotification(User $user, string $token)
+    {
+        $this->notify(new MailSuccessRegistrationNotification($user, $token));
```
lenchvolodymyr: such things better do in action. It looks like user sends email itself, but email should be send by some action. So better do $user->notify(new MailSuccessRegistrationNotification($user, $token))

backend/app/Notifications/MailSuccessRegistrationNotification.php
```diff
@@ -0,0 +1,66 @@
...
+    {
+        return ['mail'];
+    }
+
+    /**
+     * Get the mail representation of the notification.
+     *
+     * @param mixed $notifiable
+     * @return \Illuminate\Notifications\Messages\MailMessage
+     */
```
lenchvolodymyr: get rid of generated comments

backend/app/Notifications/MailSuccessRegistrationNotification.php
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+
+    /**
+     * Get the array representation of the notification.
+     *
+     * @param mixed $notifiable
+     * @return array
+     */
+    public function toArray($notifiable)
```
lenchvolodymyr: and method as well, as you do not use it

backend/resources/views/vendor/notifications/email.blade.php
```diff
@@ -0,0 +1,63 @@
+@component('mail::message')
+{{-- Greeting --}}
```
lenchvolodymyr: it this template really necessary, I did not see you use it

PR #240, opened 2019-08-24
add action for count all visitors

naytis:
https://trello.com/c/5L0bxdlV/247-add-action-that-returns-count-of-all-visitors-for-table-grouped-by-country-geo-location


naytis:
> I think everything is in order. The controller method __invoke, as I understand it, should be expanded later?

Yep. Just sequentially add execution of new actions. And pass responses from them to collect method with keys that represents in [toArray](https://github.com/BinaryStudioAcademy/bsa-2019-metrica/pull/240/files#diff-3128233a3894773c2ff3c8127167bfbdR16) method.

naytis:
> I think everything is in order. The controller method __invoke, as I understand it, should be expanded later?

PR #241, opened 2019-08-24
add geo location table component

naytis:


frontend/src/components/dashboard/geo_location/GroupedTable.vue
```diff
@@ -0,0 +1,133 @@
...
+                    { text: '', align: 'center', value: 'visitors' },
+                    { text: '', align: 'center', value: 'new_visitors' },
+                    { text: '', align: 'center', value: 'sessions' },
+                    { text: '', align: 'center', value: 'bounce_rate' },
+                    { text: '', align: 'center', value: 'avg_session_time' },
+                ],
+            };
+        },
+        computed: {
+            tableData () {
```
TatsuUkraine: This computed prop is redundant, unless you going to change data format, instead you can just pass items to table component

frontend/src/plugins/router.js
```diff
@@ -104,10 +105,11 @@ export default new Router({
                                             },
                                         },
                                         {
-                                            path: 'geo-locations',
-                                            name: 'geo-locations',
+                                            path: 'geo-location',
+                                            name: 'geo-location',
+                                            component: GeoLocation,
                                             meta: {
```
LymarMaryna: In routes it is also necessary to remove meta, I think.

PR #242, opened 2019-08-24
Feature/dashboard page

wyyi:


frontend/src/pages/Dashboard.vue
```diff
@@ -0,0 +1,15 @@
+<template>
+    <ContentLayout title="Dashboard">
+        dashboard text
+    </ContentLayout>
+</template>
+<script>
+    import ContentLayout from "../components/layout/ContentLayout";
```
devhackerone: import ContentLayout from '@/components/layout/ContentLayout.vue';
It might be better to use the full path to the component, so it will always remain correct when you move this file

frontend/src/pages/Dashboard.vue
```diff
@@ -0,0 +1,15 @@
...
+    export default {
+        name: 'Dashboard',
+        components: { ContentLayout }
```
devhackerone: I think it's still worth adding the component of adding a site in its absence
mixins: [isWebsite],

frontend/src/pages/Dashboard.vue
wyyi: didn't notice that, autocomplete)

frontend/src/pages/Dashboard.vue
```diff
@@ -0,0 +1,15 @@
+<template>
+    <ContentLayout title="Dashboard">
+        dashboard text
```
devhackerone: remove

PR #254, opened 2019-08-25
Feature/add echo service

yashuk803:


frontend/src/services/storage.js
```diff
@@ -24,6 +25,10 @@ class Storage {
         return this.set(tokenKeyName, token);
     }
 
+    getTokenType() {
+        return this.get(tokenTypeName);
```
TatsuUkraine: I'm not sure that we store token type in storage. So if you actually want to move token type somewhere to const, first of all you need to replace it's definition everywhere. But in general it could be just enough to out it in config

PR #257, opened 2019-08-26
Add map list component

LymarMaryna:


frontend/src/components/dashboard/geo_location/MapList.vue
```diff
@@ -0,0 +1,84 @@
...
+            }
+        },
+        computed: {
+            listData () {
+                if (!this.dataItems.length) {
+                    return [];
+                }
+                let totalSum = 0;
+
+                this.dataItems.map((item) => {
```
TatsuUkraine: If you need just to loop through all items, better to use forEach, map is for transformation

frontend/src/components/dashboard/geo_location/MapList.vue
```diff
@@ -0,0 +1,84 @@
...
+    </VContainer>
+</template>
+<script>
+    export default {
+        props: {
+            dataItems: {
+                type: Array,
+                required: true,
+            },
+            parameter: {
```
TatsuUkraine: What parameter means? It's better to give more specific name to prop, to describe what this prop is for

PR #259, opened 2019-08-26
Feature/add endpoint for button unique page views

SerhiiPlotnikov:


backend/app/Repositories/EloquentButtonDataPageViews.php
```diff
@@ -7,15 +7,34 @@
 use App\Contracts\Common\DatePeriod;
 use App\Entities\Visit;
 use App\Repositories\Contracts\ButtonDataPageViews;
+use Illuminate\Support\Facades\DB;
 
 class EloquentButtonDataPageViews implements ButtonDataPageViews
```
lenchvolodymyr: I asked @KoR82 to move this repository to another namespace, so it is better to merge after that fix

backend/app/Repositories/EloquentButtonDataPageViews.php
```diff
@@ -7,15 +7,34 @@
...
 {
```
lenchvolodymyr: Move this to EloquentButtonDataRepository

PR #269, opened 2019-08-27
Feature/add piecharts devices widget for displaying os and device

devhackerone:


frontend/src/components/widgets/PieChartItem.vue
```diff
@@ -0,0 +1,87 @@
...
+                    width: 110,
+                    height: 110,
+                    pieHole: 0.87,
+                    pieSliceBorderColor: 'none',
+                    legend: 'none',
+                    pieSliceText: 'none',
+                    tooltip: {
+                        trigger: 'hover'
+                    },
+                    slices: Object.assign({}, this.data.map(item => {
```
TatsuUkraine: why you need Object.assign here?

also it's probably better make own computed prop for slices, just to keep it simpler

frontend/src/pages/Dashboard.vue
```diff
@@ -43,6 +50,12 @@
             ...mapActions('dashboard', {
                 fetchingActivityChartData: FETCHING_ACTIVITY_CHART_DATA,
             }),
-        }
+        },
+        data () {
```
TatsuUkraine: it seems you not using local state, so it can be removed

frontend/src/store/modules/devices/actions.js
```diff
@@ -0,0 +1,34 @@
...
+    SET_DATA_FETCHING,
+    RESET_DATA_FETCHING
+} from "./types/mutations";
+
+import { getTimeByPeriod } from "@/services/periodService";
+import { fetchDevicesAndSystemsData } from "@/api/widgets/devicesAndSystemsService";
+
+export default {
+    [CHANGE_SELECTED_PERIOD]: (context, period) => {
+        context.commit(SET_SELECTED_PERIOD, period);
```
TatsuUkraine: you may need to dispatch `FETCH_WIDGET_INFO` action when period is changes

frontend/src/store/modules/widgets/state.js
```diff
@@ -0,0 +1,16 @@
+export default {
```
TatsuUkraine: why this state is needed?

frontend/src/pages/Dashboard.vue
wyyi: it was for some other component, didn't not notice it was not being used

PR #271, opened 2019-08-27
Feature/pageviews-endpoint-chart-avg-time-on-page-by-daterange

KoR82:


backend/app/Actions/Visits/GetPageViewsAvgTimeAction.php
```diff
@@ -0,0 +1,30 @@
...
+
+    public function execute(GetPageViewsAvgTimeRequest $request): ButtonValue
+    {
+        try {
+            $websiteId = Auth::user()->website->id;
+        } catch (\Exception $exception) {
+            throw new WebsiteNotFoundException();
+        }
+
+        return new ButtonValue((string)$this->repository->getAvgTimeOnPageBetweenDate($request->period(), $websiteId));
```
lenchvolodymyr: if string is too long, it is better to put value in the variable

PR #274, opened 2019-08-28
Add endpoint to get visitors bounce rate

LymarMaryna:


backend/app/Actions/Visitors/GetVisitorsBounceRateByParameterAction.php
```diff
@@ -0,0 +1,146 @@
...
+    public function __construct(TableVisitorsRepository $tableVisitorsRepository)
+    {
+        $this->tableVisitorsRepository = $tableVisitorsRepository;
+    }
+
+    public function execute(GetVisitorsBounseRateByParameterRequest $request): GetVisitorsBounseRateByParameterResponce
+    {
+        $parameter = $request->parameter();
+        switch ($parameter) {
+            case 'city':
```
TatsuUkraine: Consider use const, in that way you will have all variants defined in one place. For example in this particular example, you have same string duplicate 3 times

backend/app/Contracts/Visitors/VisitorsBounceRateByParameter.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+namespace App\Contracts;
+
+interface VisitorsBounceRateByParameter
```
TatsuUkraine: Not quite sure if you need interface, it seems you have only one implementation, and always will have only one. (ignore this comment if it's common practice in project))

backend/app/Actions/GeoLocation/GetGeoLocationItemsAction.php
```diff
@@ -79,7 +79,7 @@ public function execute(GetGeoLocationItemsRequest $request): GetGeoLocationItem
                 $item['all_visitors_count'],
                 $item['new_visitors_count'] ?? 0,
                 $item['all_sessions_count'],
-                $item['bounced_visitors_count'] ?? 0 / $item['all_visitors_count'] ?? 1,
+                isset($item['bounced_visitors_count'])? $item['all_visitors_count'] / $item['bounced_visitors_count'] : 0,
```
lenchvolodymyr: the relation should be vice versa
$item['bounced_visitors_count'] / $item['all_visitors_count']

and put it to variable, because the line is too long

backend/app/Contracts/Visitors/VisitorsBounceRateByParameter.php
lenchvolodymyr: Yes, it is not required to create interface for one implementaion. We have some general interfaces for visitors api, because there should be some contract for responses, as they all similar. It is not required to get rid of that, but pay attention to this comment in future

PR #282, opened 2019-08-28
Fix list data, when selected bounce rate parameter

LymarMaryna:


frontend/src/components/dashboard/geo_location/MapList.vue
```diff
@@ -44,20 +44,30 @@
...
-                });
+                if(this.displayedParameter === 'bounce_rate') {
+                    return this.dataItems.map((item) => {
+                        return {
+                            country: item.country,
+                            percentage: Math.round(Number(item[this.displayedParameter])*100)
+                        };
+                    });
+                } else {
+                    let totalSum = 0;
```
TatsuUkraine: no need to wrap all this block in else statement, if already has return, so if you unwrap this block nothing will change, but it will be cleaner

PR #286, opened 2019-08-28
Feature/send notification visits

yashuk803:


backend/app/Actions/Visits/CreateVisitAction.php
```diff
@@ -88,6 +90,9 @@ public function execute(CreateVisitRequest $request): void
         $visit->geo_position_id = $geoPosition->id;
 
         $this->visitRepository->save($visit);
+
+        $user = $visit->visitor->website->user;
+        Notification::send($user, new NewVisitsNotification($visit));
```
lenchvolodymyr: 1) it is better to create event "VisitCreated" and dispatch it here with $visit. VisitCreated::dispatch($visit)
2) Then add event listener and register it. In your event listener send notification to private channel with websiteId. Because, if we will have teams, all team should see changes on widget, not only user that has website.

https://laravel.com/docs/5.8/events

backend/app/Notifications/NewVisitsNotification.php
```diff
@@ -0,0 +1,65 @@
...
+     * @return array
+     */
+    public function via($notifiable)
+    {
+        return ['broadcast'];
+    }
+
+    public function broadcastOn()
+    {
+        return new PrivateChannel('active-users.' . $this->visit->page->website_id);
```
lenchvolodymyr: It is good

backend/app/Notifications/NewVisitsNotification.php
```diff
@@ -0,0 +1,65 @@
...
+     *
+     * @return void
+     */
+    public function __construct(Visit $visit)
+    {
+        $this->visit = $visit;
+    }
+
+    /**
+     * Get the notification's delivery channels.
```
lenchvolodymyr: get rid of comments

backend/routes/channels.php
```diff
@@ -11,6 +11,12 @@
...
 */
 
+use Illuminate\Support\Facades\Broadcast;
+
 Broadcast::channel('App.User.{id}', function ($user, $id) {
     return (int) $user->id === (int) $id;
 });
+
+Broadcast::channel('active-users.{website_id}', function ($user, $website_id) {
+    return $user->website->id == $website_id;
```
lenchvolodymyr: always use strict comparison "==="

backend/tests/Feature/VisitsApiTest.php
```diff
@@ -302,6 +305,19 @@ public function testCreateVisitAction()
             ->json('POST', $url, $data, $headers)
             ->assertStatus(200);
 
+        $visit = Visit::latest()->first();
+        $user = $this->user;
```
lenchvolodymyr: it should be different tests

PR #302, opened 2019-08-29
feature/link-page-views-table-with-backend

mes1901:
https://trello.com/c/uo9BYbcd


frontend/src/store/modules/page_views/actions.js
```diff
@@ -16,4 +32,20 @@ export default {
...
     },
+    [FETCH_PAGE_VIEWS_TABLE_DATA]: (context) => {
+        context.commit(SET_IS_FETCHING);
+
+        const period = getTimeByPeriod(context.state.selectedPeriod);
+
+        return fetchTableValues(period.startDate.unix(), period.endDate.unix())
+            .then(response => {
+                context.commit(SET_PAGE_VIEWS_TABLE_DATA, response);
+                context.commit(RESET_IS_FETCHING);
```
TatsuUkraine: no need to duplicate this commit twice - use just `finally` instead of catch, and perform this commit there

frontend/src/components/dashboard/page_views/GroupedTable.vue
```diff
@@ -60,18 +55,19 @@
...
-                    { text: '', align: 'center', value: 'avg_time' },
+                    { text: '', align: 'center', value: 'count_page_views' },
                 ],
             };
         },
         computed: {
-            tableData () {
-                return this.items;
+            getTableData () {
+                return this.items.map((item) => {
```
TatsuUkraine: it seems this computed prop is useless, you can just use prop `items` in teemplate

PR #304, opened 2019-08-29
Feature/link page views buttons with backend

SerhiiPlotnikov:


frontend/src/store/modules/page_views/actions.js
```diff
@@ -16,4 +40,52 @@ export default {
...
+            button.time.startDate.unix(),
+            button.time.endDate.unix()
+        ).then(response => {
+            let payload = {
+                buttonType: button.type,
+                value: response.value
+            };
+            context.commit(SET_BUTTON_VALUE, payload);
+            context.commit(RESET_BUTTON_FETCHING, button.type);
+        }).catch(error => {
```
TatsuUkraine: use finally and commit `RESET_BUTTON_FETCHING` once there

frontend/src/store/modules/page_views/actions.js
```diff
@@ -16,4 +40,52 @@ export default {
             context.commit(RESET_BUTTON_FETCHING, data.button);
         }
     },
+    [FETCH_PAGE_DATA]: (context) => {
+        const time = getTimeByPeriod(context.state.selectedPeriod);
```
TatsuUkraine: not quite sure that you need to define time for next two actions here, each of this actions can do this on they own

PR #323, opened 2019-08-30
Feature/subscribe channel new visit notification

yashuk803:


frontend/src/store/modules/dashboard/actions.js
```diff
@@ -49,63 +52,28 @@ export default {
...
-       });
+        context.commit(SET_BUTTON_FETCHING);
 
-       const result = [];
-        items.forEach((element) => {
-            if(result.length > 0) {
-                if(!result.find( (item => item.url === element.url && item.visitorId === element.visitorId))) {
+        return getActivityDataItems().then(response => {
+            response.sort( (a, b) => {
+                return  a.date - b.date || a.url - b.url || a.visitor - b.visitor;
```
lenchvolodymyr: are you sure you can subtract urls?

frontend/src/store/modules/dashboard/actions.js
```diff
@@ -49,63 +52,28 @@ export default {
...
+            });
+
+            const result = [];
+            response.forEach((element) => {
+                if(result.length > 0) {
```
lenchvolodymyr: in empty array you also won't find anything, so I think this condition doesn't have any sense

frontend/src/components/dashboard/home/ActiveVisitorsCard.vue
```diff
@@ -82,12 +82,22 @@
             gradientDirection: 'left',
             polling: null,
         }),
+        mounted() {
+            const channel = echoInstance.private('private-active-users.'+ this.website.id);
+            channel.listen('VisitCreated', function(data) {
+                console.log(data);
```
lenchvolodymyr: I am not sure you need console.log)

frontend/src/store/modules/dashboard/actions.js
```diff
@@ -114,9 +80,18 @@ export default {
...
+            moment().diff(moment(item.date), 'minutes') < 5
         );
         context.commit(SET_ACTIVITY_DATA_ITEMS, data);
     },
 
+    [REFRESH_ACTIVITY_DATA_ITEMS]: (context, data) => {
+        const items =  context.state.activityData.items;
+        const index = _.indexOf(items, _.find(items, {url: data.url, visitor: data.visitor}));
+        if(index >= 0) {
+            items.splice(index, 1);
```
lenchvolodymyr: I am not sure you can mutate state inside the action. You should do it inside the mutator I suppose. @TatsuUkraine am I right?

TatsuUkraine: Yep

TatsuUkraine: I would also replace index searching approach with findIndex method

TatsuUkraine: or event replace it with filter and spread operator

PR #327, opened 2019-08-31
get from backend data for fetchingActivityChartData

yashuk803:


frontend/src/components/dashboard/home/ActiveVisitorsCard.vue
```diff
@@ -106,20 +102,29 @@
         },
         created() {
             this.fetchingActivityDataItems();
+            this.fetchingActivityChartData();
```
TatsuUkraine: Please rename both methods to appropriate one, from the names of those methods it seems that you have recursive call

frontend/src/store/modules/dashboard/actions.js
```diff
@@ -108,8 +112,34 @@ export default {
...
+            const result = [];
+            if(response.length > 0) {
+                arrayOfDates.map((item) => {
+                    const value = response.find(x =>
+                        moment(moment(x.date, "DD/MM/YYYY H:mm:ss")).unix() === item.unix()
+                    );
+                    if(value) {
+                        result.push(value.value);
+                    } else {
+                        result.push(0);
```
TatsuUkraine: Could you explain why you need 0 in result array?

frontend/src/store/modules/dashboard/actions.js
```diff
@@ -108,8 +112,34 @@ export default {
...
+                    }
+                });
+            }
+            context.commit(SET_ACTIVITY_CHART_DATA, result);
+        }).catch((response) => {
```
TatsuUkraine: This catch us redundant

PR #328, opened 2019-08-31
add visits density by hour and day widget

naytis:
https://trello.com/c/per9imIR/299-add-widget-for-displaying-user-by-time-of-the-day
https://trello.com/c/hRUwMZDv/291-visitors-api-for-chart-of-page-views-accepts-as-period-milliseconds-instead-of-seconds


backend/database/factories/VisitFactory.php
```diff
@@ -7,13 +7,14 @@
...
 use App\Entities\Visitor;
+use Carbon\Carbon;
 use Faker\Generator as Faker;
 
 $factory->define(Visit::class, function (Faker $faker) {
     $session = Session::inRandomOrder()->first();
 
     return [
-        'visit_time' => $session->start_session,
+        'visit_time' => $session->start_session->addHours(rand(0, 100))->toDateTimeString(),
```
lenchvolodymyr: random in factory is supposed to break some test, at least tests that use this factory and do not set visit_time explicitly become unpredictable after these changes. If you need range of visit_time, set it in your test explicitly.

backend/composer.lock
```diff
@@ -60,16 +60,16 @@
         },
         {
             "name": "aws/aws-sdk-php",
-            "version": "3.110.0",
```
lenchvolodymyr: As you did not add any of libraries return back old composer.lock

frontend/src/components/dashboard/home/VisitsDensityWidget.vue
```diff
@@ -0,0 +1,216 @@
...
+            v-if="isFetching"
+        />
+        <div class="widget-title">
+            Users by time of day
+        </div>
+        <VueApexCharts
+            type="heatmap"
+            height="350"
+            :options="chartOptions"
+            :series="drawHeatmap()"
```
TatsuUkraine: Better to use computed prop, method won't be cached

frontend/src/components/dashboard/home/VisitsDensityWidget.vue
```diff
@@ -0,0 +1,216 @@
...
+        computed: {
+            ...mapGetters('visits_density_widget', {
+                getSelectedPeriod: GET_SELECTED_PERIOD,
+                getVisitsData: GET_VISITS_DATA,
+                isFetching: IS_FETCHING
+            }),
+        },
+        created() {
+            this.fetchWidgetData();
+            this.drawHeatmap();
```
TatsuUkraine: Any reason why you call this method on component created?

frontend/src/components/dashboard/home/VisitsDensityWidget.vue
```diff
@@ -0,0 +1,216 @@
...
+        </div>
+        <VueApexCharts
+            type="heatmap"
+            height="350"
+            :options="chartOptions"
+            :series="drawHeatmap()"
+            class="visits-heatmap"
+        />
+        <PeriodDropdown
+            style="margin-left: 10px;"
```
TatsuUkraine: Don't use inline style, you SFC has separate tag for that

frontend/src/components/dashboard/home/VisitsDensityWidget.vue
```diff
@@ -0,0 +1,216 @@
...
+            this.drawHeatmap();
+        },
+        methods: {
+            ...mapActions('visits_density_widget', {
+                changeSelectedPeriod: CHANGE_SELECTED_PERIOD,
+                fetchWidgetData: FETCH_WIDGET_DATA
+            }),
+            drawHeatmap () {
+                if (_.isEmpty(this.getVisitsData)) {
+                    this.chartOptions.xaxis.labels.show = false;
```
TatsuUkraine: Not quite sure that method, which is responsible for providing data, should do anything with local state (at least it seems this method is used just as data provider)

frontend/src/components/dashboard/home/VisitsDensityWidget.vue
```diff
@@ -0,0 +1,216 @@
...
+
+                let series = [];
+
+                for (let hour = 0; hour < 24; hour++) {
+                    let row = {};
+                    row.name = this.hours[hour] || '';
+                    row.data = [];
+
+                    for (let day = 0; day < 7; day++) {
+                        let visitsData = (_.find(this.getVisitsData, { 'hour': hour, 'day': day }));
```
TatsuUkraine: It looks like items is an array, array has methods like `find`, so you probably no need to use lodash here

frontend/src/components/dashboard/common/PeriodDropdown.vue
```diff
@@ -1,38 +1,38 @@
...
-                items: [
-                    {title: 'Today', value: period.PERIOD_TODAY},
-                    {title: 'Yesterday', value: period.PERIOD_YESTERDAY},
-                    {title: 'Last 7 days', value: period.PERIOD_LAST_WEEK},
-                    {title: 'Last month', value: period.PERIOD_LAST_MONTH},
-                    {title: 'Last quartal', value: period.PERIOD_LAST_QUARTAL},
-                    {title: 'All period', value: period.PERIOD_ALL}
-                ]
-            };
+            items: {
```
TatsuUkraine: Not quite sure if you need to define items as a prop here, it's better to do this in local state, or in computed prop, if you want to make it immutable for sure (not a blocker)

frontend/src/components/dashboard/common/VisitsDensityPeriodDropdown.vue
```diff
@@ -0,0 +1,41 @@
...
+        name: "VisitsDensityPeriodDropdown",
+        components: {
+            DefaultDropdown
+        },
+        props: {
+            value: {
+                type:String,
+                required: true
+            },
+            items: {
```
TatsuUkraine: Not quite sure if you need to define items as a prop here, it's better to do this in local state, or in computed prop, if you want to make it immutable for sure (not a blocker)

PR #331, opened 2019-08-31
add page timings fields to visit entity, update visit creation processing

TumkoVolodymyr:
https://trello.com/c/NVH2XklL


backend/database/migrations/2019_08_30_200942_add_timing_fields_to_visits_table.php
```diff
@@ -0,0 +1,36 @@
...
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('visits', function (Blueprint $table) {
+            $table->integer('page_load_time')->nullable()->default(450);
+            $table->integer('domain_lookup_time')->nullable()->default(150);
```
lenchvolodymyr: domain lookup time fronted won't send. The only server can make this request and calculate this information.

lenchvolodymyr: And it shouldn't have default, and others property as well. Default will break statistic

backend/app/Actions/Visits/CreateVisitAction.php
```diff
@@ -88,6 +88,9 @@ public function execute(CreateVisitRequest $request): void
         $visit->page_id = $page->id;
         $visit->visitor_id = $visitor->id;
         $visit->geo_position_id = $geoPosition->id;
+        $visit->page_load_time = $request->getPageLoadTime();
+        $visit->server_response_time = $request->getServerResponseTime();
+        $visit->domain_lookup_time = $request->getDomainLookupTime();
```
lenchvolodymyr: For server response time and dns lookup add methods, and make requests in order to calculate this data.

backend/app/Utils/HostIndicationsProvider.php
```diff
@@ -0,0 +1,52 @@
+<?php
+
+
+namespace App\Utils;
+
+class HostIndicationsProvider
```
lenchvolodymyr: Rename it to service (HostIndicationsService), get rid of static methods, and use laravel service container instead. I mean inject service to constructor of your action

backend/app/Http/Requests/AbstractChartHttpRequest.php
```diff
@@ -0,0 +1,47 @@
...
+
+declare(strict_types=1);
+
+namespace App\Http\Requests;
+
+use App\Http\Request\ApiFormRequest;
+use App\Rules\Timestamp;
+use App\Rules\TimestampAfter;
+
+abstract class AbstractChartHttpRequest extends ApiFormRequest
```
lenchvolodymyr: I like your idea to move general logic to some abstraction. But instead of inheritance, it is much better to use traits

PR #337, opened 2019-09-01
Feature/systems and devices endpoint

wyyi:


frontend/src/api/widgets/devicesAndSystemsService.js
```diff
@@ -1,15 +1,24 @@
...
+const url = config.getApiUrl();
 
 const fetchDevicesAndSystemsData = (startDate, endDate) => {
-    return requestService.get(resourceUrl + '/widget/devices', {}, {
+    const params = {
         'filter[startDate]': startDate,
         'filter[endDate]': endDate
-    }).then(response => devicesAndSystemsTransformer(response.data.devices))
+    };
+    return requestService.get(url + '/devices/stats', {}, params)
```
lenchvolodymyr: you can use Promise.all here

```
return Promise.all([
   requestService.get(url + '/devices/stats', {}, params),
   requestService.get(url + '/os/most-popular', {}, params)
]).then(([ devicesResponse , systemsResponse  ]) => {
     // ...
})
```

TatsuUkraine:
It seems yes, in general need to be careful with combining API resources in single API service methid

PR #342, opened 2019-09-01
rid button sign in sign up, remove variable with in chart

yashuk803:


frontend/src/components/layout/header/TopNavbar.vue
```diff
@@ -11,33 +11,14 @@
             </span>
             <span>Metrica</span>
         </VToolbarTitle>
-        <UserControls v-if="isLoggedIn" />
```
TatsuUkraine: It will remove user info from header completely, not sure that we need to do this

yashuk803: Oh my Lord! You are right! Returned.

PR #343, opened 2019-09-02
add script for check visitor and visit

LymarMaryna:


frontend/scripts/metrica.js
```diff
@@ -0,0 +1,231 @@
...
+        routes: {
+            create_visitor: '/visitors',
+            create_visit: '/visits'
+        }
+    };
+
+    const getTrackingId = () => {
+        let scripts = document.getElementsByTagName('script');
+        let index = scripts.length - 1;
+        let myScript = scripts[index];
```
lenchvolodymyr: are you sure that script will be the last on a page? Probably it is better to use querySelector() or find by a value of source, something like this "[src=https://metrica.fun/script.js?]". Or, as it also will be on the stage we can put to instruction set tracking number globally, during the initialization.

window._metricaTracking = {
      trackingNumber: <>,
      startTime: Date.now()
};

frontend/scripts/metrica.js
```diff
@@ -0,0 +1,231 @@
+(() => {
+    let state = {
+        host: 'http://stage.metrica.fun/api/v1',
```
lenchvolodymyr: I think it should be inside .env file, but leave it like this for a while, until I setup minifier and babel, as you use ES6 we need to compile it for more ancient browsers.

frontend/scripts/metrica.js
```diff
@@ -0,0 +1,231 @@
...
+    const getTrackingId = () => {
+        let scripts = document.getElementsByTagName('script');
+        let index = scripts.length - 1;
+        let myScript = scripts[index];
+
+        return (myScript.src.split('tracking_id' + '=')[1] || '').split('&')[0]
+    };
+
+    const createVisitor = (tracking_number) => {
+        return fetch(state.host + state.routes.create_visitor, {
```
lenchvolodymyr: create over "fetch" some function, because we probably will need to change implementaion from fetch to more generic, unless the babel manages it

frontend/scripts/metrica.js
```diff
@@ -0,0 +1,231 @@
...
+        });
+    };
+
+    const getUserDevice = () => {
+        let device,
+            find,
+            userAgent;
+        device = {};
+
+        window.device = device;
```
lenchvolodymyr: for global variables use our initial object, like window._metricaTracking, that should be initiated at the same time the script is being started. Actually, the user integrated tracking should initiate that object.

frontend/scripts/metrica.js
```diff
@@ -0,0 +1,231 @@
...
+
+        userAgent = window.navigator.userAgent.toLowerCase();
```
lenchvolodymyr: be more accurate ((windiw.navigator || {}).userAgent || '').toLowerCase(); Or better create helper, like getProperty(window, 'navigator.userAgent', ''), that will parse by dot and retrieve safely property. And ideally it would be better to check type, as number for instance doesn't have toLowerCase method, and js will throw exception

frontend/scripts/metrica.js
```diff
@@ -0,0 +1,231 @@
...
+
+        return device;
+    };
+
+    const getUserAgent = () => {
+        return window.navigator.userAgent;
+    };
+
+    const setToken = (token) => {
+        localStorage.setItem('visitor_token', token);
```
lenchvolodymyr: localStorage not always accessible, see as it is done inside the platform metrica. There should be some storageService

lenchvolodymyr: and wrap such things with try catch, in order not to have undesirable exceptions

frontend/scripts/metrica.js
```diff
@@ -0,0 +1,231 @@
...
+        } else if(getUserDevice().tablet()) {
+            device = 'tablet'
+        } else if(getUserDevice().mobile()) {
+            device = 'mobile'
+        }
+        return device;
+    };
+
+    const getResolutionWith = () => {
+        return screen.width;
```
lenchvolodymyr: https://caniuse.com/#search=screen

property screen is not supported by some browsers, be careful

frontend/scripts/metrica.js
LymarMaryna: This was a temporary solution, as there is no script buildyet, we cannot use the value from .env

frontend/scripts/metrica.js
```diff
@@ -0,0 +1,269 @@
...
+                localStorage.setItem('test-local-storage', 1);
+                localStorage.removeItem('test-local-storage');
+                return 'localStorage';
+            } catch (e) {
+                return 'sessionStorage';
+            }
+        },
+        getTrackById() {
+            let myScript = document.getElementById('metrica');
+            return (myScript.src.split('tracking_id' + '=')[1] || '').split('&')[0];
```
lenchvolodymyr: Probably regexp would be better

frontend/scripts/metrica.js
```diff
@@ -0,0 +1,269 @@
...
+        },
+        getUserAgent() {
+            return Helper.getPropByString(window, 'navigator.userAgent');
+        },
+        setToken(token) {
+            try {
+                this.storage().setItem('visitor_token', token);
```
lenchvolodymyr: For names that are global add prefix _metrica

frontend/scripts/metrica.js
```diff
@@ -0,0 +1,269 @@
...
+            try {
+                localStorage.setItem('test-local-storage', 1);
+                localStorage.removeItem('test-local-storage');
+                return 'localStorage';
+            } catch (e) {
+                return 'sessionStorage';
+            }
+        },
+        getTrackById() {
+            let myScript = document.getElementById('metrica');
```
lenchvolodymyr: metrica is not unique name, and some sites can use this id for their needs and then it will break a logic. Query selector still better, or id should be 100% unique

frontend/scripts/metrica.js
```diff
@@ -0,0 +1,269 @@
...
+            try {
+                localStorage.setItem('test-local-storage', 1);
+                localStorage.removeItem('test-local-storage');
+                return 'localStorage';
+            } catch (e) {
+                return 'sessionStorage';
+            }
+        },
+        getTrackById() {
+            let myScript = document.querySelector("script[src^='http://stage.metrica.fun/metrica.js?']");
```
lenchvolodymyr: okay, but divide in the state host and api suffix. Then in query use state.host + '/metrica.js'.

and retrieve from src tracking_number with regular expression. something like this /tracking_id=(\d+)/i, but should be checked

frontend/src/components/website/TrackWebsite.vue
```diff
@@ -1,15 +1,15 @@
 <template>
     <pre>
     <code>
-    &lt;!-- Global site tag (gtag.js) - Google Analytics --&gt;
-    &lt;script async src='https://www.googletagmanager.com/gtag/js?id={{ trackingNumber }}'&gt;&lt;/script&gt;
+    &lt;!-- Global site tag (metrica.js) - Metrica Analytics --&gt;
+    &lt;script async src='https://stage.metrica.fun/metrica.js?tracking_id={{ trackingNumber }}'&gt;&lt;/script&gt;
```
lenchvolodymyr: take it from .env
https://stage.metrica.fun

frontend/scripts/metrica.js
```diff
@@ -0,0 +1,309 @@
...
+
+            find = (needle) => {
+                return userAgent.indexOf(needle) !== -1;
+            };
+
+            return device;
+        }
+    };
+    try {
+        window.onload = () => {
```
lenchvolodymyr: move handler to function (e.g. onDomReady). Then check if document.readyState !== 'complete' listen onload event, unless call function immediately. Because your script can be loaded both before page has loaded or after

frontend/scripts/metrica.js
```diff
@@ -0,0 +1,309 @@
...
+            if (token) {
+                this.createVisit();
+            } else {
+                this.createVisitor(this.getTrackById())
+                    .finally(() => this.createVisit());
+            }
+
+        },
+        setObjectMetricaConf(object, prop) {
+            if(Helper.isObjectMetricaConfig(object, prop)) {
```
lenchvolodymyr: better take explicitly _metricaTrackingConfig from "window" object. And if it is an array then check if it has startDate property and tracking number. If not then it means your script has loaded earlier then script on the page was executed. It is possible when script was retrieved from browser cache. In this case you should fill config yourself, retrieving tracking number from script tag.

frontend/scripts/metrica.js
```diff
@@ -0,0 +1,319 @@
...
+            };
+
+            find = (needle) => {
+                return userAgent.indexOf(needle) !== -1;
+            };
+
+            return device;
+        }
+    };
+    const onDomReady = () => {
```
lenchvolodymyr: not exactly, if document.readyState === 'complete' is true that means page has already been loaded, it means window.onload will have never been called. But, you need to initialize your application anyway. So  onDomReady should have only initialization:

```
const onDomReady = () => {
                    window._metricaTracking.endTime = new Date();
                    window._metricaTracking.initialize();
                };

if (document.readyState !== 'complete') { 
     window.onload = onDomReady;
} else {
     onDomReady();
}
````

File: frontend/scripts/metrica.js [hunk 16]
```diff
@@ -0,0 +1,319 @@
...
+                    .finally(() => this.createVisit());
+            }
+
+        },
+        setObjectMetricaConf() {
+            let metricaConfig = Helper.getPropByString(window, '_metricaTrackingConfig');
+            if(metricaConfig === undefined) {
+                window._metricaTrackingConfig = [];
+            }
+            if(metricaConfig[0][0] !== 'dateStart' || metricaConfig[0][1] === undefined) {
```
lenchvolodymyr: if condition above is true, this string will break script:

metricaConfig = undefined;
metricaConfig[0] - Uncaught exception

make it safety

frontend/package-lock.json
```diff
@@ -3440,7 +3440,8 @@
       "version": "4.6.0",
```
lenchvolodymyr: replace package-lock with file from develop, as you did not add any packages

frontend/scripts/metrica.js
```diff
@@ -0,0 +1,321 @@
...
+                metricaConfig.push('tracking_id', this.getSearchParams('tracking_id'));
+            }
+
+            this.configMetrica = metricaConfig;
+
+        },
+        getSearchParams(k){
+            let params = {};
+            let myScript = document.querySelector(`script[src^='${state.host}metrica.js?']`);
+            myScript.src.replace(/[?&]+([^=&]+)=([^&]*)/gi,(s,k,v) => {params[k] = v});
```
lenchvolodymyr: Jesus, why so complicated

$result = (src || '').match(/tracking_id=(\d+)/i);

if ($result === null) throw new Error('Tracking id not found');

return $result[1];

frontend/scripts/metrica.js
```diff
@@ -0,0 +1,321 @@
...
+                this.createVisit();
+            } else {
+                this.createVisitor(this.getTrackById())
+                    .finally(() => this.createVisit());
+            }
+
+        },
+        setObjectMetricaConf() {
+            let metricaConfig = Helper.getPropByString(window, '_metricaTrackingConfig');
+            if(metricaConfig === undefined) {
```
lenchvolodymyr: probably, if metricaConfig === undefined, you should set metricaConfig = {};

test-site/about.html
```diff
@@ -4,6 +4,15 @@
     <meta charset="UTF-8">
     <title>About</title>
     <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=0">
+    <script id="metrica" async src="http://stage.metrica.fun/metrica.js?tracking_id=00000148" ></script>
+    <script>
+        window._metricaTrackingConfig = window._metricaTrackingConfig || [];
```
lenchvolodymyr: you are defining _metricaTrackingConfig as array, but above you work with it as with object.

test-site/about.html
```diff
@@ -4,6 +4,15 @@
     <meta charset="UTF-8">
     <title>About</title>
     <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=0">
+    <script id="metrica" async src="http://stage.metrica.fun/metrica.js?tracking_id=00000148" ></script>
```
lenchvolodymyr: https://

frontend/scripts/metrica.js
```diff
@@ -0,0 +1,321 @@
...
+                window._metricaTrackingConfig = [];
+            }
+            if(!('dateStart' in metricaConfig) || metricaConfig['dateStart'] === undefined) {
+                metricaConfig.push('dateStart', new Date());
+            }
+
+            if(!('tracking_id' in metricaConfig) || metricaConfig['tracking_id'] === undefined) {
+                metricaConfig.push('tracking_id', this.getSearchParams('tracking_id'));
```
lenchvolodymyr: below your accessing to metricaConfig as a two-dimensional array, but here you put values as in a one-dimensional array.

PR #351, opened 2019-09-02
Feature/teams-endpoints-invite-and-delete-team-member

KoR82:


frontend/src/components/dashboard/home/VisitsDensityWidget.vue
```diff
@@ -0,0 +1,205 @@
...
+                        show: false
+                    },
+                    plotOptions: {
+                        heatmap: {
+                            radius: 6,
+                        }
+                    },
+                    states: {
+                        hover: {
+                            filter: {
```
lenchvolodymyr: it is not your changes, please check in develop is it correct

backend/app/Actions/Teams/InviteTeamMemberAction.php
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    public function execute(InviteTeamMemberRequest $request): void
+    {
+        $website = $this->websiteRepository->getById($request->websiteId());
+        $password = '';
+        $teamMember = $this->userRepository->getByEmail($request->email());
+
+        if (!$teamMember) {
+            $password = Str::random(8);
```
lenchvolodymyr: I suppose, there should be a whole registration process because the user also should confirm an account. Use RegisterAction here

frontend/src/components/dashboard/home/VisitsDensityWidget.vue
KoR82: the same is in develop

backend/app/Actions/Teams/InviteTeamMemberAction.php
KoR82: Fixed

PR #358, opened 2019-09-03
add page timings vue page component and route

TumkoVolodymyr:
https://trello.com/c/ahEpcYAL


frontend/src/pages/PageTimings.vue
```diff
@@ -0,0 +1,16 @@
+<template>
```
lenchvolodymyr: It would be better to move page under directory SpeedOverview, or something like that, in order to reduce trash inside directory

frontend/src/plugins/router.js
```diff
@@ -137,10 +139,18 @@ export default new Router({
...
-                                        title: 'Speedoverview'
-                                    },
+                                    component: SpeedOverview,
+                                    children: [
+                                        {
+                                            path: 'page-timings',
+                                            name: 'page-timings',
+                                            component: PageTimings,
+                                            meta: {
+                                                title: 'Page views'
```
TatsuUkraine: This is redundant

frontend/src/components/layout/Sidepanel.vue
```diff
@@ -93,7 +93,13 @@
                 {
                     icon: '/assets/icons/speed.svg',
                     text: 'Speed Overview',
-                    route: 'speedoverview'
+                    route: 'page-timings',
```
TatsuUkraine: This key probably redundant, since sidenav not using it when sublinks are defined

frontend/src/pages/PageTimings.vue
```diff
@@ -0,0 +1,16 @@
...
+    <ContentLayout />
```
TatsuUkraine: Not quite sure if you need to use content layout twice, since parent page also uses it (ensure that it won't cause any layout issues

TumkoVolodymyr: the same structure used for  visitors, I think we can try delete this component in future with same real components on this page

TatsuUkraine: Same structure doesn't mean that it's right structure)

PR #360, opened 2019-09-03
Feature/add endpoint teams

mes1901:
https://trello.com/c/lvYvToIl


backend/app/Repositories/Teams/EloquentTeamRepository.php
```diff
@@ -4,17 +4,15 @@
...
 {
     public function getTeamMembers(int $websiteId): Collection
     {
-        return DB::table('users')
-            ->websites()
-            ->wherePivot('website_id', '=', $websiteId)
+        return Website::find($websiteId)->users()
             ->wherePivot('role', '=', 'member')
             ->select('name', 'email')
             ->get();
```
KoR82: instead of "select('name','email')->get()" you can use: get(['name','email']).
And select "member id"

PR #378, opened 2019-09-04
Feature/speedoverview table endpoints

wyyi:


backend/app/Repositories/PageViews/EloquentTableDataRepository.php
```diff
@@ -0,0 +1,73 @@
...
+            ->get();
+
+        return $res->groupBy('session.system.browser')
+            ->map(function($item, $key) use ($value) {
+                return new SpeedOverviewTableValue(
+                    'browser',
+                    $key,
+                    floatval($item->average($value))
+                );
+            })->values();
```
TumkoVolodymyr: need add sorting by avg value

PR #380, opened 2019-09-04
Feature/speedoverview table component

wyyi:


frontend/src/components/dashboard/page_timings/PageTimingsTable.vue
```diff
@@ -0,0 +1,174 @@
...
+            fluid
+        >
+            <VCol>
+                <VSelect
+                    class="option-select"
+                    prefix="Show"
+                    :items="options"
+                    flat
+                    v-model="selected"
+                    @change="changeSelect"
```
TatsuUkraine: it's quite bad practice to use model event alongside with `v-model`, also having local state, that basically affect on parents data can lead  to inconsistent state, when your dropdown will show value, that will be totally different from value that was used for grouping

frontend/src/components/dashboard/page_timings/PageTimingsTable.vue
```diff
@@ -0,0 +1,174 @@
...
+    <VContainer
+        p-0
+        class="users-table"
+    >
+        <VRow
+            class="header my-3"
+            fluid
+        >
+            <VCol>
+                <VSelect
```
TatsuUkraine: maybe you can use default dropdown that we have in common folder? (if not, it's ok)

frontend/src/components/dashboard/page_timings/PageTimingsTable.vue
```diff
@@ -0,0 +1,174 @@
...
+
+<script>
+    export default {
+        name: 'GroupedTable',
+        props: {
+            items: {
+                type: Array,
+                required: true
+            },
+            option: {
```
TatsuUkraine: what `option` means?

frontend/src/components/dashboard/page_timings/PageTimingsTable.vue
```diff
@@ -0,0 +1,174 @@
...
+                type: String,
+                required: true
+            }
+        },
+        data () {
+            return {
+                selected: 'browser',
```
TatsuUkraine: it seems this local state is redundant, since you need to get current selected group value from Vuex

frontend/src/components/dashboard/page_timings/PageTimingsTable.vue
wyyi: ran of out of ideas of how to name it, it's the parameter we want to receive i.e. page_load_time/domain_response..., but couldn't call it 'parameter' since it's reserved for the filter[parameter] and that would be confusing

frontend/src/components/dashboard/page_timings/PageTimingsTable.vue
wyyi: i thinks it's a different dropdown, the default one is for time period this one is for the parameter to sort values by

TatsuUkraine: Drop-down that I was referring is placed in components/common it receives items as option

frontend/src/components/dashboard/page_timings/PageTimingsTable.vue
TatsuUkraine: `label` or `title` maybe?

frontend/src/components/dashboard/page_timings/PageTimingsTable.vue
wyyi: i'm gonna stick with this one, the default dropdown contains styling that we use for the period dropdown, this one's different

frontend/src/components/dashboard/page_timings/PageTimingsTable.vue
```diff
@@ -0,0 +1,177 @@
...
+                ],
+                headers: [
+                    { text: '', align: 'center', value: 'name' },
+                    { text: '', align: 'center', value: 'value' },
+                ],
+            };
+        },
+        methods: {
+            changeSelect () {
+                this.$emit('change', this.value);
```
TatsuUkraine: It won't work, you need to emit value, that this method receives, but the value, which you got from props

frontend/src/pages/speed_overview/PageTimings.vue
```diff
@@ -88,29 +100,30 @@
                 currentActiveButton: GET_ACTIVE_BUTTON,
                 getSelectedPeriod: GET_SELECTED_PERIOD,
                 chartData: GET_LINE_CHART_DATA,
+                tableData: GET_TABLE_DATA,
+                getGroupedParameter: GET_GROUPED_PARAMETER
             }),
+            label() {
+                const type = this.currentActiveButton;
+                return this.buttons[type]['title'];
```
TatsuUkraine: for future use `.title` instead of `['title']` when it's possible

PR #392, opened 2019-09-04
Feature/line chart component improvements

TumkoVolodymyr:
https://trello.com/c/f2uGBX38


frontend/src/store/modules/visitors/getters.js
```diff
@@ -36,8 +38,11 @@ export default {
     },
     [GET_TABLE_DATA_FETCHING]: (state) => state.tableData.isFetching,
     [GET_GROUPED_PARAMETER]: (state) => state.tableData.groupedParameter,
-    [GET_LINE_CHART_ITEMS]: (state) => state.chartData.items,
     [GET_LINE_CHART_FETCHING]: (state) => state.chartData.isFetching,
     [GET_LINE_CHART_DATA]: (state) => state.chartData,
     [FETCH_TABLE_DATA]: (state) => state.tableData,
+    [GET_FORMAT_LINE_CHART_DATA]: (state) => {
+        state.chartData.items.sort((a, b) => a.date - b.date);
```
lenchvolodymyr: sort changes array itself, so it mutates state. First you need to make a copy, second sort it:

const items = [...state.chartData.items].sort(/*...*/)

TatsuUkraine: Also take in account, that it seems we have at least 3 places with same issue, so ensure that in other places it also will be fixed

PR #415, opened 2019-09-06
Feature/add dropdown menu for website selection

mes1901:
https://trello.com/c/BQe1pPjK


backend/app/Actions/Website/GetCurrentUserWebsiteAction.php
```diff
@@ -15,14 +15,14 @@ public function __construct(WebsiteRepository $repository)
         $this->repository = $repository;
     }
 
-    public function execute(int $websiteId): GetCurrentUserWebsiteResponse
+    public function execute(string $websiteId): GetCurrentUserWebsiteResponse
```
lenchvolodymyr: Why did you do it? This method works with int, it means you have to cast any value to int before you pass to action

backend/app/Http/Controllers/Api/WebsiteController.php
```diff
@@ -45,7 +45,7 @@ public function update(string $id, EditWebsiteHttpRequest $request, EditWebsiteA
         return ApiResponse::success(new WebsiteResource($response->getWebsite()));
     }
 
-    public function getCurrentUserWebsite(int $websiteId): ApiResponse
+    public function getCurrentUserWebsite(string $websiteId): ApiResponse
```
lenchvolodymyr: The same, method should work with int value

backend/routes/api.php
```diff
@@ -155,6 +155,7 @@
         ], function () {
             Route::get('/', 'TeamController@getTeam');
             Route::post('/', 'TeamController@inviteTeamMember');
+            Route::get('/relate', 'TeamController@getRelateUserWebsites');
```
lenchvolodymyr: relate is verb, you cannot use verbs in routes. Even you are getting websites as I understood, so it should be resource /websites/me, or smth like this

frontend/src/store/modules/website/actions.js
```diff
@@ -24,9 +31,14 @@ export default {
...
         const id = context.state.currentWebsite.id;
+
         return getCurrentUserWebsite(id).then(response => {
             context.commit(SET_CURRENT_WEBSITE, response.data);
-        }).catch(() => {
+            if(id === 0) {
+                context.commit(SET_DEFAULT_SELECTED_WEBSITE, response.data);
+            }
+        })
+            .catch(() => {
```
lenchvolodymyr: Something happend with indetion

backend/app/Actions/Website/GetCurrentUserWebsiteAction.php
mes1901: It was an errors in tests

frontend/src/store/modules/website/actions.js
```diff
@@ -118,5 +129,26 @@ export default {
...
+    },
+
+    [CHANGE_SELECTED_WEBSITE]: (context, payload) => {
+        if (context.state.relateUserWebsites.selectedWebsite.value === payload.value) {
+            return;
+        }
+
+        context.commit(SET_SELECTED_WEBSITE, payload);
+        context.dispatch(FETCH_CURRENT_WEBSITE)
+            .catch(error => {
```
lenchvolodymyr: This catch will never be called, as inside action you call catch and don't throw error further

frontend/src/components/layout/header/WebsitesDropdown.vue
```diff
@@ -0,0 +1,52 @@
...
+                type:String,
+                required: true
+            },
+        },
+        created() {
+            this.fetchRelateWebsites();
+        },
+        computed: {
+            ...mapGetters('website', {
+                websites: GET_RELATE_WEBSITES
```
TatsuUkraine: Don't spread get, fetch and change actions across application. Fetch probably should be done when user authorized. Get and change actions should be done in same component, otherwise it going to be hard to maintain

frontend/src/store/modules/website/actions.js
```diff
@@ -118,5 +129,23 @@ export default {
...
-    }
+    },
+
+    [CHANGE_SELECTED_WEBSITE]: (context, payload) => {
+        if (context.state.relateUserWebsites.selectedWebsite.value === payload.value) {
+            return;
+        }
+
+        context.commit(SET_SELECTED_WEBSITE, payload);
+        context.dispatch(FETCH_CURRENT_WEBSITE);
```
TatsuUkraine: If you want chain promise to make some actions on error on success, action should return result from dispatch

frontend/src/store/modules/website/actions.js
```diff
@@ -118,5 +130,25 @@ export default {
...
         context.commit(RESET_WEBSITE_DATA);
-    }
+    },
+
+    [CHANGE_SELECTED_WEBSITE]: (context, payload) => {
+        if (context.state.relateUserWebsites.selectedWebsite.value === payload.value) {
+            return;
+        }
+
+        context.commit(SET_SELECTED_WEBSITE, payload)
```
TatsuUkraine: Commit doesn't return promise

frontend/src/components/layout/header/UserControls.vue
```diff
@@ -71,54 +86,73 @@
...
+                showErrorMessage: SHOW_ERROR_MESSAGE
             }),
             endSession() {
                 this.logout();
                 this.resetData();
                 this.$router.push({ name: 'home' });
-            }
+            },
+            changeWebsite(data) {
+                this.changeSelectedWebsite(data)
```
TatsuUkraine: This won't work, sine your action doesn't return anything

frontend/src/store/modules/website/actions.js
```diff
@@ -24,10 +31,15 @@ export default {
     [FETCH_CURRENT_WEBSITE]: (context) => {
         context.commit(SET_FETCH_TRUE);
         const id = context.state.currentWebsite.id;
+
         return getCurrentUserWebsite(id).then(response => {
             context.commit(SET_CURRENT_WEBSITE, response.data);
-        }).catch(() => {
+            if(id === 0) {
```
TatsuUkraine: What is this? Why you ever fetch something with id === 0?

mes1901:
@TatsuUkraine We have a certain task in trello which performs vue-routes access

frontend/src/store/modules/website/actions.js
mes1901: We have to return current website id to backend for getting data. After user login there is no current website in the state. That's why we return 0. At the backend if id=0 we return  website where user is owner or first existing where user is member

TatsuUkraine: It's totally wrong, if you pass 0, endpoint should return website with id 0. Also, that is why I said, that everything that related to website fetching should be revisited

TatsuUkraine:
@mes1901 in this particular PR you trying to hook into existing store, without trying to refactor it, which is totally wrong, I don't know what route access ticket describes, but this particular changes touches access guard, when user doesn't have website, so it should be fixed here too

if you want to add multiple websites on a frontend, you need to change the way how websites are fetching in a first place. Also since there going to be more than 1 website, then current implementation with current website fetching after login becomes redundant

frontend/src/store/modules/website/actions.js
mes1901: in this context the check is for setting selected item in drop down menu as default

TatsuUkraine: Again, you fetching data by ID === 0!

mes1901:
there cannot be such a case when the user doesn't have a site. the user can be either the owner or member.

TatsuUkraine:
I can give you at least 2 cases when user doesn't have website:
- user just registered
- user was removed from from member list by owner

frontend/src/api/transformers.js
```diff
@@ -0,0 +1,11 @@
+const relateWebsites = (item) => {
```
TatsuUkraine: It seems this function not in use

frontend/src/components/layout/header/UserControls.vue
```diff
@@ -71,54 +85,82 @@
...
                 this.$router.push({ name: 'home' });
+            },
+            changeWebsite(data) {
+                let website = {};
+                this.websites.map((item) => {
+                    item.domain === data.value ? website = item : website;
+                });
+                this.changeSelectedWebsite(website);
+            },
+            getWebsites() {
```
TatsuUkraine: It's better to make this function computed property

frontend/src/components/layout/header/UserControls.vue
```diff
@@ -71,54 +85,82 @@
...
+                return this.websites.map((item) => {
+                    return {
+                        title: item.name + ' - ' + item.role,
+                        value: item.domain,
+                    };
+                });
+            },
+            getSelectedValue() {
```
TatsuUkraine: You can make getter in vuex for this, or you can move this function to computed prop

frontend/src/components/layout/header/UserControls.vue
```diff
@@ -71,54 +85,82 @@
...
                 this.logout();
                 this.resetData();
                 this.$router.push({ name: 'home' });
+            },
+            changeWebsite(data) {
+                let website = {};
+                this.websites.map((item) => {
+                    item.domain === data.value ? website = item : website;
+                });
+                this.changeSelectedWebsite(website);
```
TatsuUkraine: Not sure that all this logic should happen here, your action CHANGE_SELECTED_WEBSITE can receive just ID and then you can define selected website based on it inside vuex

frontend/src/store/modules/website/mutations.js
```diff
@@ -14,41 +17,41 @@ export default {
...
-            tracking_number: ''
-        };
+        state.relateUserWebsites = [];
+        state.selectedWebsite = undefined;
         state.isCurrentWebsite = false;
         state.isFetchedWebsite = false;
-    }
+    },
+    [SET_SELECTED_WEBSITE]: (state, website) => {
+        state.selectedWebsite = website;
```
TatsuUkraine: I'm not sure sure if you need to make a snapshot of selected website. You can simply store ID and in GET_SELECTED_WEBSITE find website object based on it

PR #417, opened 2019-09-06
Fix/seed with different websites and roles

KoR82:


lenchvolodymyr:
How is it posdible?

lenchvolodymyr:
Why are there so many changes?

KoR82:
> How is it posdible?

What exactly possible? We need some users with different roles and websites for testing swithing between websites.

KoR82:
> Why are there so many changes?

I see it in "feature/teams" only and can't understand why. 
Steps I went through for this result:
- git checkout -b feature/teams
- git pull origin feature/teams
- git checkout -b fix/seed-with-different-websites-and-roles
- git push origin fix/seed-with-different-websites-and-roles

May be my IDE make these things.

lenchvolodymyr:
If you do it from develop 

git checkout -b feature/teams

then probably it throws error that branch already exists, as key -b means create new branch and switch to it, so just to switch branch you need to call checkout without -b

lenchvolodymyr:
Merge develop in feature/teams, because I cannot see changes

KoR82:
> If you do it from develop
> 
> git checkout -b feature/teams
> 
> then probably it throws error that branch already exists, as key -b means create new branch and switch to it, so just to switch branch you need to call checkout without -b

Of course. I tried to show "all steps" before creating "feature/teams". For the next branches I used checkout without "-b"

KoR82:
> Merge develop in feature/teams, because I cannot see changes

Done. My changes in this PR are 2 commits only:

https://github.com/BinaryStudioAcademy/bsa-2019-metrica/pull/417/commits/d0b86667efc55b2e8f5088557c17dcec6ef1e61e

https://github.com/BinaryStudioAcademy/bsa-2019-metrica/pull/417/commits/854f06143457ea8de375506a6b0d794c881a1ed9

lenchvolodymyr:
I still see changes of all team, I'd like to look only at your changes, fix it. It is not appropriate to see commits, and anyway there is a typo 'ueser'

PR #419, opened 2019-09-06
add units display for bounce rate

wyyi:


frontend/src/components/common/LineChart.vue
```diff
@@ -38,6 +39,10 @@
                 type: Boolean,
                 required: true,
             },
+            units: {
+                type: String,
+                required: true
```
TatsuUkraine: Probably better to make it optional with `default: ''`

PR #424, opened 2019-09-08
improve visitors flow diagram

naytis:
https://trello.com/c/ryHU3Nr3/442-get-rid-of-animation-on-the-behavior-diagram-when-next-step-is-produced
https://trello.com/c/JQGAl3fu/443-when-click-on-transition-line-highlight-all-path

### DIFF
--- a/frontend/src/components/dashboard/visitors_flow/VisitorsFlowDiagram.vue
+++ b/frontend/src/components/dashboard/visitors_flow/VisitorsFlowDiagram.vue
@@ -35,6 +35,7 @@
                     '3rd Interaction'
                 ],
                 height: 600,
+                tooltip: {},
                 nodes: [
                     { id: 1, name: 'United States' },
                     { id: 2, name: 'United Kingdom' },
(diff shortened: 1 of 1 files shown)


frontend/src/components/dashboard/visitors_flow/VisitorsFlowDiagram.vue
```diff
@@ -86,20 +88,31 @@
...
 
+            createExits() {
+                this.exits = [];
+                let lastNodeId = this.nodes.slice(-1)[0].id;
+
+                for (let sourceInd = 5; sourceInd < lastNodeId; sourceInd++) {
+                    let sourceId = this.nodes[sourceInd].id;
+                    let value = Math.floor(Math.random() * 50);
+                    let exit = { source: sourceId, value: value, index: sourceId - 1 };
+                    this.exits.push(exit);
```
TatsuUkraine: All keys in local state are reactive, so each push will cause props evaluation. It's better to define variable, and put in local state at the end

PR #434, opened 2019-09-09
show data only for exist values

TumkoVolodymyr:
https://trello.com/c/9u8TprJ4


backend/app/Actions/Visitors/BounceRateAction.php
```diff
@@ -32,8 +32,11 @@ public function execute(BounceRateRequest $request): BounceRateResponse
         $collection = new Collection();
         do {
             $all = $allVisitorsByTimeFrameValues[$start]??0;
+            if ($all === 0) {
```
lenchvolodymyr: filtering of empty values is not good approach, as it will show a falsy result. Better trimming zero result from both sides start and end

PR #436, opened 2019-09-09
Feature/add refetch data when selected another website

mes1901:
https://trello.com/c/ki2YX5vV


frontend/src/mixins/isChangeSelectedWebsite.js
```diff
@@ -0,0 +1,19 @@
+import store from "../store";
+
+export const isChangeSelectedWebsite = {
+    mounted() {
+        store.watch(
```
TatsuUkraine: Define watcher with component, but not with store. Mixins allows you to define exact same properties as in regular component

frontend/src/mixins/isChangeSelectedWebsite.js
```diff
@@ -0,0 +1,19 @@
...
+            (state) => state.selectedWebsite,
+            (newValue, oldValue) => {
+                if(newValue !== oldValue) {
+                    this.onWebsiteChange();
+                }
+            }
+        );
+    },
+    methods: {
+        onWebsiteChange: function () {
```
TatsuUkraine: No need to define method inside mixin, it should be inside component. And inside watcher you need just check if method exists before you call it

PR #438, opened 2019-09-09
Feature/change new website save data

mes1901:
https://trello.com/c/7fANfW0b


frontend/src/store/modules/website/mutations.js
```diff
@@ -20,11 +21,12 @@ export default {
...
     },
-    [UPDATE_CURRENT_WEBSITE]: (state, website) => {
-        state.currentWebsite = {
-            ...state.currentWebsite,
-            ...website
-        };
+    [UPDATE_CURRENT_WEBSITE]: (state, name) => {
+        state.relateUserWebsites.map(website => {
+            if(website.id === state.selectedWebsite) {
+                website.name = name;
```
TatsuUkraine: Don't update object like this, it can break reactivity. Instead - find needed object, and out new array instance in store

PR #439, opened 2019-09-09
fix calculation of bounce rate data for table

TumkoVolodymyr:
https://trello.com/c/i2RayoNh


backend/app/Actions/Visitors/GetVisitorsBounceRateByParameterAction.php
```diff
@@ -101,10 +95,11 @@ public function execute(GetVisitorsBounceRateByParameterRequest $request): GetVi
...
         $response = $collection->map(function ($item) use ($parameter) {
             $bounce_rate = isset($item['bounced_visitors_count']) ? $item['bounced_visitors_count'] / $item['visitors_count'] : 0;
-            return new BounceRateVisitors(
+            return new TableValue(
                 $parameter,
-                $item[$parameter],
-                $bounce_rate
+                is_array($item[$parameter])?$item[$parameter][0]:$item[$parameter],
+                (string)($item['bounced_visitors_count']??0),
+                $bounce_rate * 100
```
lenchvolodymyr: we have agreed to return from backend values 0 - 1 and transform it to percents on frontend

frontend/src/api/visitors/bounceRateService.js
```diff
@@ -39,7 +39,7 @@ const fetchChartValues = (startDate, endDate, interval) => {
 };
 
 const fetchTableValues = (startDate, endDate, groupBy) => {
-    return requestService.get(resourceUrl + '/table-visitors/count-total', {}, {
+    return requestService.get(resourceUrl + '/table-visitors/bounce-rate', {}, {
```
lenchvolodymyr: you changed route here, but did not in api.php, is it correct?

TumkoVolodymyr: yes, someone set wrong route here and it is right route

backend/app/Actions/Visitors/GetVisitorsBounceRateByParameterAction.php
TumkoVolodymyr: I did as there is for all other table data route

PR #440, opened 2019-09-10
Add script for collect errors

LymarMaryna:


frontend/scripts/metrica.js
```diff
@@ -9,12 +9,40 @@
             });
         }
     }
+
+    window.onerror = (errorMsg, errorUrl) => {
```
lenchvolodymyr: I am not sure, but probably addEventListener better, in order not to be changed by user

And fifth parameter is error object, and you can take error.stack from here, because line number is not exactly what you need

https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onerror

frontend/scripts/metrica.js
```diff
@@ -9,12 +9,40 @@
...
+        let data = {
+            message: errorMsg,
+            stack_trace: errorUrl,
+            page: window._metricaTracking.getPage(),
+            page_title: window._metricaTracking.getTitle(),
+
+        };
+
+        if (token) {
+            headers = {
```
lenchvolodymyr: You can do it like this

{
                ...headers,
                'x-visitor': 'Bearer ' + token
            }

PR #443, opened 2019-09-10
Feature/link menu with backend displaying due to access rights

mes1901:
https://trello.com/c/FIojHJFW


frontend/src/components/layout/Sidepanel.vue
```diff
@@ -57,80 +57,107 @@
...
+                    }
+                });
+
+                return [
+                    {
+                        icon: '/assets/icons/home.svg',
+                        text: 'Home',
+                        route: 'dashboard'
+                    },
+                    {
```
TatsuUkraine: maybe it's better to hide section entirely if there is no sublinks available? just a thought, since such items won't do anything on click

PR #449, opened 2019-09-10
add showing successful message on recovery password page

SerhiiPlotnikov:


frontend/src/store/modules/auth/actions.js
```diff
@@ -64,7 +64,8 @@ export default {
 
     [RESET_PASSWORD]: (context, payload) => {
         return resetPassword(payload).then(() => {
-                return `Your reset password link was created. Check your email ${payload.email}, please.`;
+                return ` Check your inbox. We just sent a link to reset your password via email to ${payload.email}.
```
lenchvolodymyr: Oh gosh, do not return html from the action, do it inside a component. You can even wrap this text inside component and triggering it by some flag dependent on the result

PR #453, opened 2019-09-11
Fix/realtime activity card

yashuk803:


frontend/src/store/modules/dashboard/actions.js
```diff
@@ -103,10 +118,12 @@ export default {
                     result.push(0);
                 }
             });
+
             const chunk = (arr, size) =>
                 arr.reduce((acc, _, i) => (i % size)
```
lenchvolodymyr: looks scary, there is a method chunk in lodash library

PR #456, opened 2019-09-11
Feature/add protect vue routes

mes1901:
https://trello.com/c/fx2uQYS9

### DIFF
--- a/frontend/src/plugins/router.js
+++ b/frontend/src/plugins/router.js
@@ -1,5 +1,6 @@
 import Vue from 'vue';
 import Router from 'vue-router';
+import store from '../store/index';
 import Login from '../pages/Login.vue';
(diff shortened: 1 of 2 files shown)


frontend/src/plugins/router.js
```diff
@@ -107,11 +108,21 @@ export default new Router({
                                             path: 'visitors',
                                             name: 'visitors',
                                             component: Visitors,
+                                            beforeEnter: (to, from, next) => {
```
TatsuUkraine: not quite sure that this check should happen here. We have route guard on layout itself to fetch data. this variant could cause case when this check will happen earlier than data will be fetched. Instead it's better to create mixin, and use meta to define what kind of access is needed for each particular page

PR #466, opened 2019-09-11
Fix/graph all period select

TumkoVolodymyr:
https://trello.com/c/xj2HxYRP

### DIFF
--- a/backend/app/Actions/Visits/GetBounceRateChartByDateRangeAction.php
+++ b/backend/app/Actions/Visits/GetBounceRateChartByDateRangeAction.php
@@ -7,6 +7,7 @@ namespace App\Actions\Visits;
 use App\DataTransformer\ChartValue;
 use App\Entities\Website;
 use App\Repositories\Contracts\ChartVisitRepository;
+use Carbon\Carbon;
 use Illuminate\Support\Collection;
 use Illuminate\Support\Facades\Auth;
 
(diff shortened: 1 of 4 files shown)


frontend/src/pages/Visitors.vue
```diff
@@ -138,12 +138,9 @@
                 formatLineChartData:GET_FORMAT_LINE_CHART_DATA,
             }),
             units() {
-                switch (this.currentActiveButton) {
-                case BOUNCE_RATE:
+                if (this.currentActiveButton === BOUNCE_RATE) {
```
lenchvolodymyr: Why did you delete this?

backend/app/Actions/Visits/GetBounceRateChartByDateRangeAction.php
```diff
@@ -22,6 +23,8 @@ public function __construct(ChartVisitRepository $repository)
     public function execute(GetBounceRateChartByDateRangeRequest $request): GetBounceRateChartByDateRangeResponse
     {
         $from = $request->period()->getStartDate();
+        $minStarDate = new Carbon('01/01/2018');
```
lenchvolodymyr: What is the magic date?

lenchvolodymyr: instead of '01/01/2018' probably better to take date from the very first element, doesn't it?

frontend/src/pages/Visitors.vue
TumkoVolodymyr: because on page page view we don use units for avg time, and I delete here units for consistency

backend/app/Actions/Visits/GetBounceRateChartByDateRangeAction.php
TumkoVolodymyr: yes it would be better find the earlier element, but I decided to not complicate

